Issue #15, opened 2019-06-09
Chart is not redrawn on data update


erolasan:
I am using a pie chart which gets data fed from a StreamProvider and whenever I receive the updated data the chart is not redrawn. I have tried debugging it and I can see that the updated values are set to the PieChartSectionData's, but the chart doesn't redraw itself automatically.

apaolino:
I'm in a similar situation with the difference I receive data (List<FLPlot>) from a StreamBuilder, then I pass it to the widget (line chart, same code from your example) but the chart is not redrawn

imaNNeoFighT:
I know, this is a bug, and I know what I should exactly do to fix it,
But @erolasan opened a PR to fix it and needs some changes, we should wait for it.
Be a patient, and It will be fixed ASAP, follow it here #16

imaNNeoFighT:
@apaolino fixed on 0.0.6, check it out :)

ilyesksr:
I am having the same problem, the Chart is not redrawn on data update until I hover with the mouse on the pie Chart widget (i build on Windows)

PR #16, opened 2019-06-09
enabled pie chart repainting

erolasan:
#15 enabled pie chart repainting


erolasan:
tested it with all 3 types of charts, it works correctly

imaNNeoFighT:
Thanks, man!

PR #18, opened 2019-06-11
Feature/line chart extra lines

agustka:
Adding extra line data to line charts that contains horizontal and vertical lines.

### DIFF
--- a/lib/src/chart/line_chart/line_chart_data.dart
+++ b/lib/src/chart/line_chart/line_chart_data.dart
@@ -10,10 +10,12 @@ import 'package:flutter/material.dart';
 class LineChartData extends AxisChartData {
   final List<LineChartBarData> lineBarsData;
   final FlTitlesData titlesData;
+  final ExtraLinesData extraLinesData;
 
   LineChartData({
     this.lineBarsData = const [],
     this.titlesData = const FlTitlesData(),
+    this.extraLinesData,
     FlGridData gridData = const FlGridData(),
     FlBorderData borderData,
     double minX,
@@ -37,7 +39,7 @@ class LineChartData extends AxisChartData {
   ) {
     lineBarsData.forEach((lineBarChart) {
       if (lineBarChart.spots == null || lineBarChart.spots.isEmpty) {
-        throw Exception('spots could not be null or empty');
+        throw Exception('spots must not be null or empty');
       }
     });
     if (lineBarsData.isNotEmpty) {
@@ -131,6 +133,9 @@ class LineChartBarData {
   /// to show dot spots upon the line chart
   final FlDotData dotData;
 
+  // to show line chart annotations such as average line and vertical dot lines
+  final ExtraLinesData extraLinesData;
+
   const LineChartBarData({
     this.spots = const [],
     this.show = true,
@@ -142,6 +147,7 @@ class LineChartBarData {
     this.isStrokeCapRound = false,
     this.belowBarData = const BelowBarData(),
     this.dotData = const FlDotData(),
+    this.extraLinesData = const ExtraLinesData(),
   });
 }
 
@@ -200,4 +206,52 @@ class FlDotData {
     this.dotSize = 4.0,
     this.checkToShowDot = showAllDots,
   });
+}
+
+/// This class holds data about drawing chart annotations (data decorations) such as average line and data point lines
+class ExtraLinesData {
+  final bool show;
+
+  // Average line
+  final bool showAverageLine;
+  final LineStyle averageLineStyle;
+
+  // Data point lines
+  final bool showDataPointLines;
+  final bool terminateAtChartLine;
+  final LineStyle dataPointLineStyle;
+
+  const ExtraLinesData({
+    this.show = true,
+
+    // Average line
+    this.showAverageLine = true,
+    this.averageLineStyle = const LineStyle(),
+
+    // Data point lines
+    this.showDataPointLines = true,
+    this.terminateAtChartLine = true,
+    this.dataPointLineStyle = const LineStyle(),
+  });
+}
+
+class LineStyle {
+  final Color color;
+  final double width;
+  final bool dashed;
+  final DashDefinition dashDefinition;
+
+  const LineStyle({
+    this.color = Colors.black12,
+    this.width = 1,
+    this.dashed = false,
+    this.dashDefinition = const DashDefinition(solidWidth: 1, gapWidth: 3.5)
+  });
+}
+
+class DashDefinition {
+  final double solidWidth;
+  final double gapWidth;
+
+  const DashDefinition({this.solidWidth, this.gapWidth});
 }
--- a/lib/src/chart/line_chart/line_chart_data.dart
+++ b/lib/src/chart/line_chart/line_chart_data.dart
@@ -167,7 +167,7 @@ class BelowBarData {
   final Offset gradientTo;
 
 
-  final List<VerticalLine> verticalLines;
+  final FlLine verticalLines;
 
   /// if more than one color provided gradientColorStops will hold
   /// stop points of the gradient.
@@ -208,21 +208,11 @@ class FlDotData {
   });
 }
 
-class VerticalLine extends FlLine {
+class HorizontalLine extends FlLine {
   final double x;
 
-  const VerticalLine({
-    @required this.x,
-    Color color = Colors.black,
-    double strokeWidth = 1,
-  }) : super(color: color, strokeWidth: strokeWidth);
-}
-
-class HorizontalLine extends FlLine {
-  final double y;
-
   const HorizontalLine({
-    @required this.y,
+    @required this.x,
     Color color = Colors.black,
     double strokeWidth = 1,
   }) : super(color: color, strokeWidth: strokeWidth);
--- a/lib/src/chart/line_chart/line_chart_data.dart
+++ b/lib/src/chart/line_chart/line_chart_data.dart
@@ -218,7 +218,7 @@ class HorizontalLine extends FlLine {
   }) : super(color: color, strokeWidth: strokeWidth);
 }
 
-/// This class holds data about drawing chart annotations (data decorations) such as average line and data point lines
+/// This class holds data about drawing horizontal lines at user specified locations in the chart
 class ExtraLinesData {
   final bool show;
 
--- a/lib/src/chart/line_chart/line_chart_data.dart
+++ b/lib/src/chart/line_chart/line_chart_data.dart
@@ -208,50 +208,60 @@ class FlDotData {
   });
 }
 
-/// This class holds data about drawing chart annotations (data decorations) such as average line and data point lines
-class ExtraLinesData {
-  final bool show;
-
-  // Average line
-  final bool showAverageLine;
-  final LineStyle averageLineStyle;
-
-  // Data point lines
-  final bool showDataPointLines;
-  final bool terminateAtChartLine;
-  final LineStyle dataPointLineStyle;
-
-  const ExtraLinesData({
-    this.show = true,
-
-    // Average line
-    this.showAverageLine = true,
-    this.averageLineStyle = const LineStyle(),
-
-    // Data point lines
-    this.showDataPointLines = true,
-    this.terminateAtChartLine = true,
-    this.dataPointLineStyle = const LineStyle(),
-  });
-}
-
-class LineStyle {
-  final Color color;
-  final double width;
-  final bool dashed;
-  final DashDefinition dashDefinition;
-
-  const LineStyle({
-    this.color = Colors.black12,
-    this.width = 1,
-    this.dashed = false,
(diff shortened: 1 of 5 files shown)


lib/src/chart/line_chart/line_chart_data.dart
```diff
@@ -129,6 +131,9 @@ class LineChartBarData {
   /// to show dot spots upon the line chart
   final FlDotData dotData;
 
+  // to show line chart annotations such as average line and vertical dot lines
+  final ExtraLinesData extraLinesData;
```
imaNNeoFighT: Here we don't need it, because ExtraLinesData as defined on LineChartData, we should use that to draw lines in the whole chart, and not related any LineChartBarData,
Please remove it, and for the below lines as I said we should put it on the BelowBarData, I will explain it on the next sections.

lib/src/chart/line_chart/line_chart_data.dart
```diff
@@ -198,4 +204,62 @@ class FlDotData {
     this.dotSize = 4.0,
     this.checkToShowDot = showAllDots,
   });
-}
\ No newline at end of file
+}
+
+class DashDefinition {
```
imaNNeoFighT: Please remove it, we can add line style functionality line Dash and Dot effects in next phase, here we just add two feature (Extra Lines -> you use it for avg line, and BelowBar Lines-> you use it for below bar lines), then I will add LineStyle after this PR, don't worry about that.

lib/src/chart/line_chart/line_chart_data.dart
```diff
@@ -198,4 +204,62 @@ class FlDotData {
...
+  const LineChartLine({
+    Color color,
+    double strokeWidth,
+    this.dashDefinition,
+  }) : super(color: color, strokeWidth: strokeWidth);
+}
+
+class VerticalLine extends LineChartLine {
+  final double x;
+  final double endY;
```
imaNNeoFighT: we don't need endY because ExtraLines should be drawn from y = 0 to y = height, and the only x is dynamic ( for the below line we should apply another solution), prefer this ExtraLines just for avg line.

lib/src/chart/line_chart/line_chart_data.dart
```diff
@@ -198,4 +204,62 @@ class FlDotData {
...
+    this.endY,
+    DashDefinition dashDefinition,
+    Color color = Colors.black,
+    double strokeWidth = 1,
+  }) : super(color: color, strokeWidth: strokeWidth, dashDefinition: dashDefinition);
+}
+
+class HorizontalLine extends LineChartLine {
+  final double y;
+  final double endX;
```
imaNNeoFighT: also we don't need endX, (read previous comment)

lib/src/chart/line_chart/line_chart_painter.dart
```diff
@@ -40,6 +45,8 @@ class LineChartPainter extends AxisChartPainter {
     });
 
     drawTitles(canvas, viewSize);
+
+    drawAnnotation(canvas, viewSize);
```
imaNNeoFighT: dude, this is not annotation in a generic definition, look at the problem as a generic problem, we should name it extraLines to draw extra lines on the chart, then you can use it for your problem to draw avg line, this is not related to the FL Chart.

lib/src/chart/line_chart/line_chart_painter.dart
```diff
@@ -326,4 +333,85 @@ class LineChartPainter extends AxisChartPainter {
...
+      }
+      return;
+    }
+  }
+
+  void _drawHorizontalLineLine(Canvas canvas, Size viewSize, HorizontalLine line) {
+    viewSize = getChartUsableDrawSize(viewSize);
+    extraLinesPaint.color = line.color;
+    extraLinesPaint.strokeWidth = line.strokeWidth;
+    final double x = line.endX == null ? viewSize.width : getPixelX(line.endX, viewSize);
```
imaNNeoFighT: here we can equal x with the viewSize.width every time.
because this is not for draw below lines

lib/src/chart/line_chart/line_chart_painter.dart
```diff
@@ -326,4 +333,85 @@ class LineChartPainter extends AxisChartPainter {
...
+    final double y = getPixelY(line.y, viewSize);
+    if (line.dashDefinition != null) {
+      _drawDashedLine(canvas, extraLinesPaint, 0, y, x, y, line.dashDefinition);
+    } else {
+      _drawSolidLine(canvas, extraLinesPaint, 0, y, x, y);
```
imaNNeoFighT: remove this if else and determine the line as a solid line every time ( I will add Line Style after your PR in a separate phase), and you can draw the solid line here and you don't need to put it in a new function (because there is no if else anymore)

lib/src/chart/line_chart/line_chart_data.dart
agustka: What if the user wants the line to go all the way to the top though and not necessarily end at the data point?

lib/src/chart/line_chart/line_chart_painter.dart
agustka: Residual naming from the original work when I first started. Renaming.

lib/src/chart/line_chart/line_chart_data.dart
agustka: Remove it but put it where..?

lib/src/chart/line_chart/line_chart_data.dart
agustka: All right, although I will not be able to use fl_chart library without it.

lib/src/chart/line_chart/line_chart_data.dart
agustka: Yeah but what if the user wants the line to end at a data point for example?

lib/src/chart/line_chart/line_chart_data.dart
imaNNeoFighT: As I saw you defined it to the LineChartData

agustka: Removed.

lib/src/chart/line_chart/line_chart_data.dart
imaNNeoFighT: We can not add all features in a single PR, I said multiple times that I will add it after your pR, it should add in a separate phase, because I want to reuse it on other parts such as grid and borders, this is a public library and we should implement all features generic and customizable, (for example add LineStyle on other parts).
be patient

lib/src/chart/line_chart/line_chart_data.dart
imaNNeoFighT: If we want to draw this to the specific data point then we should have start data point alongside it.
I said, use it just for full horizontal and vertical lines, and it can be used as a avg line in your usecase, then you should use BelowBarData to draw lines below dot points

lib/src/chart/line_chart/line_chart_data.dart
imaNNeoFighT: Drop it plz, if we want to make it possible to set end point then we have to add start point for it to be customizable

lib/src/chart/line_chart/line_chart_data.dart
agustka: After checking it some more I understand better what below bar data is for, so I'm dropping endY.

lib/src/chart/line_chart/line_chart_data.dart
```diff
@@ -162,6 +164,9 @@ class BelowBarData {
   final Offset gradientFrom;
   final Offset gradientTo;
 
+
+  final List<VerticalLine> verticalLines;
```
imaNNeoFighT: here we should just have a single FlLine that determines the style of line, then we can repeat it for all spot points, we don't need VerticalLine, because we know where we should draw them (below the spots)

lib/src/chart/line_chart/line_chart_data.dart
```diff
@@ -198,4 +204,37 @@ class FlDotData {
...
+
+  const VerticalLine({
+    @required this.x,
+    Color color = Colors.black,
+    double strokeWidth = 1,
+  }) : super(color: color, strokeWidth: strokeWidth);
+}
+
+class HorizontalLine extends FlLine {
+  final double y;
```
imaNNeoFighT: We have a convention on Library, 
the horizontal is all of the things that draw on the x axis,
the vertical is all of the things that draw on the y axis,
then HorizontalLine means lines that draw on the x axis and ( for example in x = 1, 2, 3, 4, 5), and y is the same  (may be it is VerticalLine in your head, but it is a convention in this library, check the FlTitlesData you can find out what does it means.)
conclusion:
we should have x here instead of y

lib/src/chart/line_chart/line_chart_painter.dart
agustka: Removed

lib/src/chart/line_chart/line_chart_painter.dart
agustka: Removed

lib/src/chart/line_chart/line_chart_data.dart
agustka: Dropped in favor of FlLine

lib/src/chart/line_chart/line_chart_data.dart
agustka: Won't this confuse everybody? A horizontal line has infinite x values... seems strange but I'll change, its your library.

lib/src/chart/line_chart/line_chart_data.dart
agustka: Dropped

lib/src/chart/line_chart/line_chart_data.dart
agustka: Y has been changed to x.

lib/src/chart/line_chart/line_chart_painter.dart
```diff
@@ -326,4 +348,31 @@ class LineChartPainter extends AxisChartPainter {
 
   @override
   bool shouldRepaint(CustomPainter oldDelegate) => false;
+
+  void drawExtraLines(Canvas canvas, Size viewSize) {
```
imaNNeoFighT: Here plz draw both horizontal and vertical lines, (firstly data class should be updated)

lib/src/chart/line_chart/line_chart_data.dart
```diff
@@ -198,4 +204,27 @@ class FlDotData {
...
+  final double x;
+
+  const HorizontalLine({
+    @required this.x,
+    Color color = Colors.black,
+    double strokeWidth = 1,
+  }) : super(color: color, strokeWidth: strokeWidth);
+}
+
+/// This class holds data about drawing chart annotations (data decorations) such as average line and data point lines
```
imaNNeoFighT: Update the comment, ( we don't need to mention anything about annotation, this word is specific for your use case)

agustka: Fixed

lib/src/chart/line_chart/line_chart_painter.dart
agustka: The drawing of the two line types has been merged into this method.

agustka: But I'm not sure what you mean with that the data class should be updated?

lib/src/chart/line_chart/line_chart_painter.dart
```diff
@@ -317,6 +323,36 @@ class LineChartPainter extends AxisChartPainter {
     }
   }
 
+  void drawExtraLines(Canvas canvas, Size viewSize) {
+    // Draw vertical data point lines under spots
```
imaNNeoFighT: her is not under spots, fix the comment

lib/src/chart/line_chart/line_chart_painter.dart
```diff
@@ -317,6 +323,36 @@ class LineChartPainter extends AxisChartPainter {
...
+    viewSize = getChartUsableDrawSize(viewSize);
+    for (LineChartBarData barData in data.lineBarsData) {
+      if (barData.belowBarData.verticalLines == null) {
```
imaNNeoFighT: why you are reading it from belowBarData?
dude draw it just like the horizontal lines,
you should have a class like this:
```
class ExtraLinesData {
  final bool show;

   // Horizontal line definitions
  final List<HorizontalLine> horizontalLines;

  // Vertical line definitions
  final List<VerticalLine> verticalLines;

   const ExtraLinesData({
    this.show = true,
    this.horizontalLines,
    this.verticalLines,
  });
}
```

both of them should define in this class

agustka: You told me to move it to below bar data...

PR #23, opened 2019-06-13
added backgroundColor (initial)

Triipaxx:
added backgroundColor property to AxisChartData which is used in the AxisChartPainter to draw a colored rect that fills the chart

BarChartData and LineChartData are passing the backgroundColor to its super class AxisChartData

### DIFF
--- a/lib/src/chart/bar_chart/bar_chart_data.dart
+++ b/lib/src/chart/bar_chart/bar_chart_data.dart
@@ -22,9 +22,11 @@ class BarChartData extends AxisChartData {
     ),
     FlBorderData borderData,
     double maxY,
+    Color backgroundColor,
   }) : super(
           gridData: gridData,
           borderData: borderData,
+          backgroundColor: backgroundColor,
         ) {
     initSuperMinMaxValues(maxY);
   }
--- a/lib/src/chart/base/axis_chart/axis_chart_data.dart
+++ b/lib/src/chart/base/axis_chart/axis_chart_data.dart
@@ -17,12 +17,16 @@ class AxisChartData extends BaseChartData {
   /// clip the chart to the border (prevent draw outside the border)
   bool clipToBorder;
 
+  /// A background color which is drawn behind th chart.
+  Color backgroundColor;
+
   AxisChartData({
     this.gridData = const FlGridData(),
     FlBorderData borderData,
     this.minX, this.maxX,
     this.minY, this.maxY,
     this.clipToBorder = false,
+    this.backgroundColor,
   }) : super(borderData: borderData);
 }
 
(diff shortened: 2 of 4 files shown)


lib/src/chart/bar_chart/bar_chart_data.dart
```diff
@@ -22,9 +22,11 @@ class BarChartData extends AxisChartData {
     ),
     FlBorderData borderData,
     double maxY,
+    Color backgroundColor,
```
imaNNeoFighT: Set a default color of Colors.transparent everywhere that you have this property ( I will mention them)

lib/src/chart/base/axis_chart/axis_chart_data.dart
```diff
@@ -17,12 +17,16 @@ class AxisChartData extends BaseChartData {
...
+  /// A background color which is drawn behind th chart.
+  Color backgroundColor;
+
   AxisChartData({
     this.gridData = const FlGridData(),
     FlBorderData borderData,
     this.minX, this.maxX,
     this.minY, this.maxY,
     this.clipToBorder = false,
+    this.backgroundColor,
```
imaNNeoFighT: Here

lib/src/chart/line_chart/line_chart_data.dart
```diff
@@ -23,10 +23,12 @@ class LineChartData extends AxisChartData {
     double minY,
     double maxY,
     bool clipToBorder = false,
+    Color backgroundColor,
```
imaNNeoFighT: And here

lib/src/chart/base/axis_chart/axis_chart_painter.dart
```diff
@@ -76,6 +80,25 @@ abstract class AxisChartPainter<D extends AxisChartData> extends BaseChartPainte
     }
   }
 
+  /// This function draws a colored background behind the chart.
+  void drawBackground(Canvas canvas, Size viewSize) {
+    if (data.backgroundColor == null) {
```
imaNNeoFighT: If we have default color, there is no need to null check, and it's okay we throw a error when user passed null.

Triipaxx:
Ok, but on the other hand, this will cause an unnecessary rendering. And transparent/translucent rendering is often more expensive over opaque rendering. I'm not sure if this is an issue with flutter/skia.

lib/src/chart/base/axis_chart/axis_chart_painter.dart
Triipaxx: Ok, but on the other hand, this will cause an unnecessary rendering. And transparent/translucent rendering is often more expensive over opaque rendering. I'm not sure if this is an issue with flutter/skia.

PR #24, opened 2019-06-13
added invertYAxis feature

Triipaxx:
added invertYAxis property to BarChartData, LineChartData and AxisChartData --> property defaults to false

LineChartPainter / BarChartPainter overrides the getTopOffsetDrawSize method to return enough vertical space for the horizontal titles that are rendered on top when y-axis is inverted
LineChartPainter / BarChartPainter fixed content rendering when y axis is inverted

AxisChartPainter inverts most of the rendering in the getPixelY method, this is done by using the calculated y value without subtracting it from the chart height
fixed grid rendering in AxisChartPainter when y axis is inverted

i've checked my changes with your samples and it looked good so far. I hope i didn't messed up your code too much. maybe we can extract some of the calculations into methods.

### DIFF
--- a/lib/src/chart/bar_chart/bar_chart_data.dart
+++ b/lib/src/chart/bar_chart/bar_chart_data.dart
@@ -23,10 +23,12 @@ class BarChartData extends AxisChartData {
     FlBorderData borderData,
     double maxY,
     Color backgroundColor,
+    bool invertYAxis = false,
   }) : super(
           gridData: gridData,
           borderData: borderData,
           backgroundColor: backgroundColor,
+          invertYAxis: invertYAxis,
         ) {
     initSuperMinMaxValues(maxY);
   }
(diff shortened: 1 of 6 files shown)


imaNNeoFighT:
I checked it, but let's keep it simple, as I explained, we just want to add top and right titles,
and no more about inverting y-axis, because it can be handled by user behind of the library,
In the library, we just add a possibility to add top titles and hide bottom titles, 
then the user can invert the y-axis by himself, and invert left titles, then give them to the chart,
I think it is okay and there is no need to make more effort on it

Triipaxx:
Ok, no Problem. I will keep my changes for now and use it on my own project till top titles will arrive.

imaNNeoFighT:
Sure,
I will add it ASAP, and make a sample of the inverted y-axis on samples :)

PR #47, opened 2019-07-12
(fix) BelowBar considers its own color stops refs #46

### LINKED ISSUE #46: Bug: BelowBarData accidentally considers BarData.colorsStop's count
When creating a `LineChartBarData` with `colors` and `colorStops` set, together with `BelowBarData`, 
`LineChartPainter.drawBelowBar` will consider `BarData`'s colorStops instead of `BelowBarData`'s.

Related method: `LineChartPainter.drawBelowBar`

daniel-v:
This fixes an issue where BelowBar ignored `gradientColorStops` property and used `colorStops` from the bar itself. Should fix #46


imaNNeoFighT:
Thanks for contribution ;)
Cheers!

Issue #100, opened 2019-11-09
Updating from 0.3.3 to 0.4.1, it breaks my charts (LineChart and PieChart)


jamesblasco:
Given a size to the father container of my chart it throws this error.

The following assertion was thrown building PieChart(duration: 150ms, dirty, dependencies: [MediaQuery], state: PieChartState#f59ef(ticker active)):
RenderBox was not laid out: RenderCustomPaint#99bfc NEEDS-LAYOUT NEEDS-PAINT 'package:flutter/src/rendering/box.dart': Failed assertion: line 1687 pos 12: 'hasSize'

imaNNeoFighT:
I will check,
Thanks

jamesblasco:
I will try to research more about it and give a proper example. It happens when i resize de view for example when i rotate the device

imaNNeoFighT:
Ok prepare a reproducers sample, It helps me to debug it faster.

jamesblasco:
I found when does it happen but don't know how to solver it. 

It happens when the chart is inside a LayoutBuilder widget. 
I replaced the _MyHomePageState of the fl_chart example with the following code and it throws the error.


```dart
class _MyHomePageState extends State<MyHomePage> {
 
  @override
  Widget build(BuildContext context) {
    return LayoutBuilder(
        builder: (BuildContext context, constraints) => Center(
            child: FractionallySizedBox(
                heightFactor: 0.5,
                widthFactor: 0.5,
                child: Scaffold(
                  body: SafeArea(
                    child: PageView(
                      children: <Widget>[
                        LineChartPage(),
                        BarChartPage(),
                        BarChartPage2(),
                        PieChartPage(),
                        LineChartPage2(),
                        LineChartPage3(),
                        LineChartPage4(),
                      ],
                    ),
                  ),
                ))));
  }
}
```

imaNNeoFighT:
Dude you are trying to show the whole page with a 0.5 scale factor.
Is it intentionally?

jamesblasco:
Yes, sorry i was testing it on ipad and i wanted to check the resize when rotating. But I think for the example is not needed

jamesblasco:
[...]

PR #110, opened 2019-11-13
Replased containerRenderBox.size to containerRenderBox.constraints.biggest

jamesblasco:
Fix [issue 100](https://github.com/imaNNeoFighT/fl_chart/issues/100). 

RenderBox.size should not be used when building the layout.

### DIFF
--- a/lib/src/chart/bar_chart/bar_chart.dart
+++ b/lib/src/chart/bar_chart/bar_chart.dart
@@ -86,7 +86,7 @@ class BarChartState extends AnimatedWidgetBaseState<BarChart> {
       },
       child: CustomPaint(
         key: _chartKey,
-        size: getDefaultSize(context),
+        size: chartSize,
         painter: BarChartPainter(
           _withTouchedIndicators(_barChartDataTween.evaluate(animation)),
           _withTouchedIndicators(showingData),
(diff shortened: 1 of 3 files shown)


lib/src/chart/bar_chart/bar_chart.dart
```diff
@@ -86,7 +86,7 @@ class BarChartState extends AnimatedWidgetBaseState<BarChart> {
       },
       child: CustomPaint(
         key: _chartKey,
-        size: getDefaultSize(context),
+        size: chartSize,
```
imaNNeoFighT: Sorry for the delay.
Please do not change this line, we set the default size it means if it has not any size constrains, it uses this default size (It will be ignored if we provide a size for it) then let it be like before, it is working properly.

lib/src/chart/line_chart/line_chart.dart
```diff
@@ -91,7 +91,7 @@ class LineChartState extends AnimatedWidgetBaseState<LineChart> {
       },
       child: CustomPaint(
         key: _chartKey,
-        size: getDefaultSize(context),
+        size: chartSize,
```
imaNNeoFighT: And here.

lib/src/chart/pie_chart/pie_chart.dart
```diff
@@ -87,7 +87,7 @@ class PieChartState extends AnimatedWidgetBaseState<PieChart> {
       },
       child: CustomPaint(
         key: _chartKey,
-        size: getDefaultSize(context),
+        size: chartSize,
```
imaNNeoFighT: And here.

imaNNeoFighT:
Helooo?

lib/src/chart/bar_chart/bar_chart.dart
jamesblasco: Okey i will remove that. I cant do it in the following days. After monday

daniel-mf:
Could we have this merged please?
Thanks guys for the good work!

PR #117, opened 2019-11-23
Gradient direction support on line chart

kamilpowalowski:
I needed a way to show low and high values on the chart so I modified `LineChartBarData` implementation to support `gradientFrom` and `gradientTo` parameters based on similar logic from `BarAreaData` implementation.
With those changes making charts like the one in attachment is possible.

Screenshot_1574536012
Simulator Screen Shot - iPhone 11 - 2019-11-24 at 12 47 27


imaNNeoFighT:
Hi @kamilpowalowski.
Good implementation, I've enjoyed it.
Thanks!

PR #148, opened 2019-12-31
allow rod not be square

yintaibing:
allow rod not be square


imaNNeoFighT:
Ok man, your solution is right and you know exactly where to fix it.
but we already have an 'isRount' property, and use it instead of defining a new `square` property.
Does it make sense?

imaNNeoFighT:
It has just been fixed in #158 with a better solution.
Thanks for contributing.

PR #150, opened 2020-01-03
respect text scale factor when drawing text

juumixx:
All painters should use the text scale factor when drawing text. Since there is no context in the chart painters available, we have to inject it during the creation of each painter.


imaNNeoFighT:
So helpful feature, and implemented well.
Thanks!

Issue #113, opened 2019-11-16
enhancement to support  horizontal bar chart.


kw2019ltd:
enhancement to support  horizontal bar and stacked bar charts.

juumixx:
I'm working on horizontal bar charts:
https://github.com/juumixx/fl_chart/commits/horizontal-bar-chart

Things that are still not behaving nicely are backdraw bars and stack items. The rest should be production ready.

ksankumar:
Any update for horizontal bar chart?

eripe970:
Any update on this?

imaNNeoFighT:
I'm busy as a bee  😢😢
I appreciate your help and your PRs

imaNNeoFighT:
We added StackedBarChart feature. I remove it from the title. Now we just need the horizontal bar chart.
Thanks!

imaNNeoFighT:
There is a temporary workaround mentioned [here](https://github.com/imaNNeoFighT/fl_chart/issues/552#issuecomment-775812719)

2shrestha22:
Any update on this?

imaNNeoFighT:
Not yet!

2shrestha22:
> There is a temporary workaround mentioned [here](https://github.com/imaNNeoFighT/fl_chart/issues/552#issuecomment-775812719)

Okay, do you have any idea about rotating TooltipIndicators,

imaNNeoFighT:
Yes, it is on the way [here](https://github.com/imaNNeoFighT/fl_chart/pull/679).

sawirricardo:
Hi, @imaNNeoFighT any updates on this? Thank you

imaNNeoFighT:
Hi @sawirricardo. Not yet sir!

spoeck:
Hi @imaNNeoFighT , is there any update about the horizontal bar chart?

imaNNeoFighT:
Hi @spoeck. 
Unfortunately, not yet!

akshatj427:
Hi @imaNNeoFighT i have recently started working with BarCharts in flutter and came across the fl_chart library as it had the highest popularity score. The current requirement for the app i am developing requires us to create a horizontal bar chart, would love to see it added to the library, is there any update on when can we expect it to be added?

imaNNeoFighT:
Hi @akshatj427 
Unfortunately, I'm not working full-time on this project (I hope I could do)
I work here in my free and leisure time. That's why everything is a little bit slow.

[...]

PR #160, opened 2020-01-09
WIP: added horizontal bar chart

juumixx:
Added horizontal bar charts.

To be able to reuse the bar_chart_data, I changed the naming of `y` and `x` to in the bars and groups to `value`, because in the horizontal bar they would be flipped.
Also, horizontal bars have the ability to put text onto the bar, which is saved in `BarChartRodData.title`.
We could also draw this text 90° rotated on the vertical bar chart to be consistent, but this is not implemented yet.

I added a temporary example here

What do you think?

PS: it would be really cool if we could format the code according to [effective dart](https://dart.dev/guides/language/effective-dart/style), I just noticed there is a lot of whitespace change in the diff which is unrelated :(


imaNNeoFighT:
Wow, there are a lot of changes.
I'm so happy to see that, it fixes #113 
I will review it ASAP, give me some time dude :)
And about the code format, I'm totally with you but do not reformat in a feature PR ever, it makes the review process hard, 
make a new PR, and do all reformat in that, I will merge it ASAP.
Thanks!

scostagabriel:
@imaNNeoFighT , thank you for this amazing library, it's really awesome!

Do you have any expectations about when to merge this PR?

berengere13:
Hello,

Have you any news about this feature ? Our team is very interested !

Thanks you very much.

imaNNeoFighT:
Hello, I think it's not gonna implement soon, 
because we are so busy, can make your hands dirty and try to implement it.
Thanks, regards!

Nazacheres:
@imaNNeoFighT Actually similar results can be achieved, just by wrapping BarChart with RotatedBox. What would be good is just enable rotation for titles and tooltips :) That's a small change.

imaNNeoFighT:
Okay, make separate issues for them to follow up.
Thanks for suggesting!

liasica:
> @imaNNeoFighT Actually similar results can be achieved, just by wrapping BarChart with RotatedBox. What would be good is just enable rotation for titles and tooltips :) That's a small change.

tooltips rotation how?

PR #171, opened 2020-01-18
Dashed lines

shamilovtim:
Screen Shot 2020-01-17 at 10 13 54 PM

The idea is to allow our different datatypes to have a `dashArray` property which will then be processed by functions in the `util.dart` file. I've started with a `pathDasher()` utility and will add a lineDasher. Hope it makes sense!


lib/src/chart/line_chart/line_chart_data.dart
```diff
@@ -236,6 +238,8 @@ class LineChartBarData {
   /// show indicators based on provided indexes
   final List<int> showingIndicators;
 
+  final CircularIntervalList<double> dashArray;
```
imaNNeoFighT: external objects should not be exposed to the user directly,
because in some scenarios we may change or remove this object,
1. maybe in the future we have our drawing effects, and we decide to not using `path_drawing` anymore.
2. maybe `path_drawing` decides to change this object, then our users are exposed to that decision.
and other scenarios.
please decide a descriptive name for that object, and let's have our properties, then in the lower layer (drawing section) we will map it to the `CircularIntervalList` and make our dashed path.
and I have an idea in my head to have a class to support other types of drawing, dashed path, dotted path, ...

shamilovtim: @imaNNeoFighT I this obj is just a circular array with a double generic. I can define the same thing in the models of this project. I assume the dependency will accept it? I'm not sure

shamilovtim:
Screen Recording 2020-01-18 at 3 17 14 PM

have some sort of bug from touching anywhere inside the chart

shamilovtim:
Bug was from floating point arithmetic. The linedasher should prefer integers.

shamilovtim:
simple array of [2,4] looks good

Screen Recording 2020-01-18 at 3 28 53 PM

shamilovtim:
killed two bird with one stone. our own CircularArray class now enforces integer list so repainting bug is gone.

lib/src/chart/line_chart/line_chart_data.dart
shamilovtim: 2ed8a8a

lib/src/utils/utils.dart
```diff
@@ -27,3 +28,14 @@ Size getDefaultSize(BuildContext context) {
 double translateRotatedPosition(double size, double degree) {
   return (size / 4) * math.sin(radians(degree.abs()));
 }
+
+/// takes in a path and returns a dathed path
+/// data.dashArray is provided
+/// dynamic because dart has no union types
+Path pathDasher(Path path, Paint painter, dynamic data) {
```
imaNNeoFighT: Why here we take dynamic?
Why we don't use the strong typed `CircularArray`?

shamilovtim:
Actually this is broken right now. The dependency does not accept my CircularArray. I think I need a type assertion

lib/src/utils/utils.dart
shamilovtim: I'll pass in the array next

shamilovtim:
How do I force the dependency to accept my different object is the same as the kind they want? In TypeScript it is simple I just use the `as` keyword.

lib/src/chart/line_chart/line_chart_data.dart
shamilovtim: there is nothing I can do about this. the type that the dependency requires is strongly typed and is not iterable which means there's nothing I can transform into it or from it. the only way around this problem is to remove the dependency and bring it inline and modify it or otherwise somehow override the class it uses. it's too much work for such a small change

shamilovtim: ok I was losing my mind but I got it. I had to cast the inner object which was just a plain List, then invoke the constructor on that casted List. Dart is soooooo strongly typed. I am used to TypeScript

shamilovtim: here we go fd0b7f3

lib/src/utils/utils.dart
shamilovtim: done

imaNNeoFighT:
You should map the custom object to the library object.
If it is done, please remove the WIP
Thanks

shamilovtim:
still WIP added dashing to FlLine. now just have to process it anywhere we `drawLine`

Screen Recording 2020-01-19 at 1 08 00 AM

shamilovtim:
now available most anywhere we draw lines

Screen Shot 2020-01-19 at 1 35 09 AM

this is now complete and probably ready to merge

imaNNeoFighT:
Well done

lib/src/chart/base/axis_chart/axis_chart_data.dart
```diff
@@ -136,19 +136,23 @@ class FlGridData {
...
-    this.color = Colors.black,
-    this.strokeWidth = 2,
-  });
+  const FlLine({this.color = Colors.black, this.strokeWidth = 2, this.dashArray});
 
   static FlLine lerp(FlLine a, FlLine b, double t) {
     return FlLine(
       color: Color.lerp(a.color, b.color, t),
       strokeWidth: lerpDouble(a.strokeWidth, b.strokeWidth, t),
+      dashArray: b.dashArray,
```
imaNNeoFighT: Please use `lerpDoubleList` (in the `lerp.dart`),
It has a better experience when we animate the chart, 
The dashes will be updated with animation.
(I know our list lerps not implemented well, I will re-implement them in the near future)

lib/src/chart/line_chart/line_chart_data.dart
```diff
@@ -264,6 +267,7 @@ class LineChartBarData {
       isStrokeCapRound: b.isStrokeCapRound,
       preventCurveOverShooting: b.preventCurveOverShooting,
       dotData: FlDotData.lerp(a.dotData, b.dotData, t),
+      dashArray: b.dashArray,
```
imaNNeoFighT: And here

lib/src/utils/utils.dart
```diff
@@ -1,6 +1,7 @@
 import 'dart:math' as math;
 
 import 'package:flutter/material.dart';
+import 'package:path_drawing/path_drawing.dart';
```
imaNNeoFighT: We don't need it anymore

lib/src/chart/base/axis_chart/axis_chart_data.dart
imaNNeoFighT: Please use `lerpDoubleList` (in the `lerp.dart`),
It has a better experience when we animate the chart, 
The dashes will be updated with animation.
(I know our list lerps not implemented well, I will re-implement them in the near future)

lib/src/utils/utils.dart
```diff
@@ -27,3 +28,26 @@ Size getDefaultSize(BuildContext context) {
 double translateRotatedPosition(double size, double degree) {
   return (size / 4) * math.sin(radians(degree.abs()));
 }
+
+/// returns a dashed path from a path
+Path generateDashedPath(Path path, List<int> dashArray) {
```
imaNNeoFighT: I highly recommend you to implement these kinds of functions with [extension methods](https://dart.dev/guides/language/extension-methods),
It's really really easy to use,
for example we can use 
`canvas.drawDashedLine(from, to, paint, arr)`
instead of 
`drawDashedLine(canvas, from, to, touchLinePaint, indicatorData.indicatorBelowLine.dashArray);`

and 
`path.toDashPath(arr)`
instead of 
`barPath = generateDashedPath(barPath, arr);`

imaNNeoFighT:
And it's ready to merge if you resolve these conversations.
Thanks for your great PRs.
I appreciate them.

lib/src/chart/base/axis_chart/axis_chart_data.dart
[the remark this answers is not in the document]
shamilovtim: Ok! 

It's an int list at this point before we cast it in the end. So here I would need to lerp int list?

shamilovtim:
Done! Feels better to use the extended class. I am still exporting the extended class and the dependency in case anyone wants to use it.  That's fine right?

lib/src/extensions/canvas_extension.dart
```diff
@@ -0,0 +1,25 @@
+import 'dart:ui';
+
+import 'package:path_drawing/path_drawing.dart';
+
+extension DashedPath on Canvas {
+  /// helper that generates dashed paths for other methods
+  void drawDashedPath(Path path, Paint painter, List<int> dashArray) {
```
imaNNeoFighT: In my opinion, it's better to have two functions,
one to convert the path to dashed path,
and the other one to draw the path(we already have this function)
so please make an extension function on the Path, and as I said before call it `toDashedPath()`, and just draw the result as a normal path (using `canvas.drawPath()` function)
My consern is applying SRP (functions should do one and only one thing), and we can reuse the `toDashedPath()` later.

lib/fl_chart.dart
```diff
@@ -1,5 +1,8 @@
 library fl_chart;
 
+export 'package:path_drawing/path_drawing.dart';
```
imaNNeoFighT: We don't need it anymore

lib/src/extensions/canvas_extension.dart
shamilovtim: It would be 3x the code to put it on path... I would need to call the generate function everywhere and do logic throughout the app

shamilovtim: 234d35e

shamilovtim:
All set?

lib/src/extensions/canvas_extension.dart
```diff
@@ -1,25 +0,0 @@
...
-      final dashedPath = dashPath(path, dashArray: CircularIntervalList<double>(castedArray));
-
-      drawPath(dashedPath, painter);
-    } else {
-      drawPath(path, painter);
-    }
-  }
-
-  /// draws a dashed line from passed in offsets
-  void drawDashedLine(Offset from, Offset to, Paint painter, List<int> dashArray) {
```
imaNNeoFighT: This is the last issue, I promise :)
please keep this one, I didn't say remove it (I said to just remove the `drawDashedPath`),
It is just fine.

lib/src/extensions/path_extension.dart
```diff
@@ -0,0 +1,24 @@
...
+      final dashedPath = dashPath(this, dashArray: CircularIntervalList<double>(castedArray));
+
+      return dashedPath;
+    } else {
+      return this;
+    }
+  }
+
+  /// get a dashed(or solid) path from two offsets
+  Path toDashedPathFromLine(Offset from, Offset to, List<int> dashArray) {
```
imaNNeoFighT: please remove it and use, `drawDashedLine` (I know it has been removed, but please bring it back).

lib/src/extensions/canvas_extension.dart
imaNNeoFighT: I know your concern, and I know we should write more codes for it,
but It helps us to maintain the code easier.

imaNNeoFighT:
Hi,
I know I'm bothering you and I'm sorry about that,
these kinds of issues, enable us to have long term maintenance.

shamilovtim:
No problem I made the two extensions

kip1980h:
Does anyone know how to make a grid

imaNNeoFighT:
Thank you :)

PR #175, opened 2020-01-21
Improve painter order

shamilovtim:
Improves the order of painting through the following:

1.  `rangeAnnotation` painting moved outside of the for loop. `extraLines` removed from the for loop except when extraLinesOnTop is enabled. probably improved performance.
2. the order of painting is revised to background → range annotations → grid → extralines → barLine → dots → (optional): extralines  → touchIndicator


lib/src/chart/line_chart/line_chart_painter.dart
```diff
@@ -86,6 +85,14 @@ class LineChartPainter extends AxisChartPainter<LineChartData>
       drawBetweenBarsArea(canvas, size, data, betweenBarsData);
```
imaNNeoFighT: It should be drawn after background and grid and ...

shamilovtim: Nice catch. What else beside the grid should it cover? I think that's it.

imaNNeoFighT: I think other things are well

shamilovtim:
good to go

shamilovtim:
I had one more commit. I needed to paint borders by calling super super paint.

lib/src/chart/line_chart/line_chart_painter.dart
```diff
@@ -812,7 +814,7 @@ class LineChartPainter extends AxisChartPainter<LineChartData>
 
         final double bottomChartPadding = getExtraNeededVerticalSpace() - getTopOffsetDrawSize();
         final Offset to =
-        Offset(getPixelX(line.x, chartUsableSize), viewSize.height - bottomChartPadding);
+            Offset(getPixelX(line.x, chartUsableSize), viewSize.height - bottomChartPadding);
```
imaNNeoFighT: Useless change.

lib/src/chart/line_chart/line_chart_painter.dart
```diff
@@ -796,7 +798,7 @@ class LineChartPainter extends AxisChartPainter<LineChartData>
 
         final double rightChartPadding = getExtraNeededHorizontalSpace() - getLeftOffsetDrawSize();
         final Offset to =
-        Offset(viewSize.width - rightChartPadding, getPixelY(line.y, chartUsableSize));
+            Offset(viewSize.width - rightChartPadding, getPixelY(line.y, chartUsableSize));
```
imaNNeoFighT: Useless change.

lib/src/chart/line_chart/line_chart_painter.dart
shamilovtim: this is from a bug in dart formatter where it had previously not formatted this line. the only way i can prevent this from occurring is to turn off dart formatter.

lib/src/chart/line_chart/line_chart_painter.dart
```diff
@@ -84,10 +83,19 @@ class LineChartPainter extends AxisChartPainter<LineChartData>
       canvas.saveLayer(Rect.fromLTWH(0, -40, size.width + 40, size.height + 40), Paint());
     }
 
+    super.superBorderPaint(canvas, size);
+    super.drawBackground(canvas, size);
+    drawRangeAnnotation(canvas, size);
```
imaNNeoFighT: Hi, I've checked it but I have a problem with it,
there is a big complexity here, I think we can simplify it.
these logics are applied only in the line chart,
and I think in the near future we need the range annotations for the bar charts,
then why not we handle it in the AxisChartPainter class, and don't touch the painting order?
we can have exact this paint order in the superclass, instead of exposing superclass paint functions,
the only drawback is moving `RangeAnnotations` class into the `AxisChartData` and handle the paint function in the `AxisChartPainter` class,
it makes breaking changes, but I think it is necessary.
What do you think?

shamilovtim: I'd say if you want to merge this and then enhance it to a more ideal state later I don't see why not but there's no reason not to merge it right now because this PR fixes bugs immediately

imaNNeoFighT: Yes, it fixes the issue (and I think it's better to have an individual issue and mention it here to track it), 
but If there is a clearly better way to do that, we should merge it in the proper way,
we know the issue, we know the immediate solution, and we know the best solution.
then please let to do it in the best solution’s way.
If you are frustrated, I'm so sorry about that, but we need to keep our code clean and readable.
I can continue it if you let me.
Thanks!

shamilovtim: Sure I have no problem with that I was just saying that the branch could be merged twice,  one to get it in production now and one to enhance it to where you want it in the future. I will rebase these commits out of my other PR.

imaNNeoFighT: No worries, 
I will do it tonight ASAP.
Thanks!

imaNNeoFighT:
I think I can't push to this PR, because it is in your fork.

shamilovtim:
Screen Shot 2020-02-06 at 2 15 13 PM

that shouldn't be the case. you don't have directions below the merge button that let you push to it?

imaNNeoFighT:
image

I think I need permission to push on your forked repo.

imaNNeoFighT:
image

I think I need permission to push on your forked repo.

shamilovtim:
I added permissions to my fork but here is what the checkbox says:
Screen Shot 2020-02-06 at 4 56 08 PM

shamilovtim:
I think you might have a duplication:
`git push shamilovtim patch-range-annotations` the name isn't shamilovtim/patch-range-annotations but only "patch-range-annotations"

shamilovtim:
also i just noticed that your current branch says 'shamilovtim-patch-range-annotations'. you could try to rename it maybe?

imaNNeoFighT:
I'm lost :))))

imaNNeoFighT:
wait for a second

shamilovtim:
in your terminal you're pushing to my repo named 'shamilovtim' a branch named shamilovtim/patch-range-annotations but your branch is actually named 'shamilovtim-patch-range-annotations' but the real branch name on my repo is only 'patch-range-annotations' 

at least i think that's what's going on. i might be wrong. lol

imaNNeoFighT:
That was a hard work :)
1. I cloned your repo, then I checked out to the `patch-range-annotations` branch
2. then I added `imaNNeoFighT/fl_chart` as a new remote (I called it`neo`)
3. then I ran `git fetch neo` command
4. cherry-pick that commit from my remote
5. rebase your branch with `neo/master`
6. force pushed to `patch-range-annotations` branch

imaNNeoFighT:
BTW, It's ready to merge in my opinion,
you can have a look at it, and approve my changes to merge.
Thanks!

shamilovtim:
I haven't tested manually but it looks fine to merge to me

PR #176, opened 2020-01-21
swapped HorizontalExtraLines, and VerticalExtraLines definitions.

imaNNeoFighT:
removed `showHorizontalLines` and `showVerticalLines` from `ExtraLinesData`


shamilovtim:
Good call

PR #184, opened 2020-01-27
Feature: Image annotation using image or svg

shamilovtim:
The purpose of this PR is to add the ability to annotate a line (whether it has a visible width or not) with an image or an icon. I have created the preliminary image annotation functionality. The user on his side needs to load the image as a future, which is standard Flutter stuff. On our side the only thing we really care is that if the image isn't null, we draw the image centered by simple center algorithm: width divided by 2 and height divided by 2. 

Screen Shot 2020-01-27 at 2 36 01 PM

Further in this PR I will add the ability to annotate using an svg rather than an image.

~This PR builds on #175 to avoid merge conflict. (Not too important since there's not much in 175).~


shamilovtim:
Example with no line (this is how we personally plan to use it):
Screen Shot 2020-01-27 at 2 51 58 PM

shamilovtim:
I have had svgs render by turning them into images and using the same image function so it looks like this is close to done

shamilovtim:
now done on horizontal lines
Screen Shot 2020-01-28 at 4 26 30 PM

shamilovtim:
Screen Shot 2020-01-28 at 10 19 23 PM

svg example using new `sizedPicture` attribute on VerticalLine and HorizontalLine. users now have the option of adding an image annotation using an image file or an svg

shamilovtim:
readme has been updated as well I believe this is ready to merge

shamilovtim:
#175 rebased out of this PR due to big changes in how that PR will be finished

shamilovtim:
@imaNNeoFighT hello, any updates on this? 🙂

imaNNeoFighT:
Hello, unfortunately, I'm as busy as bee these days, I will review it ASAP.
Sorry about the delay.
Thanks!

imaNNeoFighT:
Hi @shamilovtim, I'm thinking about it.
There is a general problem to having widgets inside the chart, it is so sophisticated.
I think we have to find a way to let the user put their own widgets inside the chart,
If it is possible then we don't need it,
you know, I'm trying to solve it in an abstract way, for example, what if the user wants to put a text instead of an image?
What do you think? can we solve it in an abstract way to put any widget inside the chart?
I think we can put our chart inside a Stack widget, then overlap some widgets upon the chart.

shamilovtim:
@imaNNeoFighT Since the inside of the chart is dynamic with trendlines that depend on data, the only realistic place to annotate the data is either on the axes themselves (this feature) or along the trendline (for example `FlSpot.y + 20`) which is not yet implemented. Beside these two options, there is very little way of knowing when a statically placed asset is going to conflict with a dynamically placed trendline without some heavy algorithms and I think it would be a waste of developer time and energy. 

Witt regards to text it can already be placed along the axes using intervals and labels. This PR allows the placement of either vectors or images along the same axes where text already has the option of existing. Beside text, images and vectors, I can't imagine there's any other kind of widget required since this encompasses everything visual (to my knowledge). 

The way I see it if someone wants an annotation along their trendline, this feature doesn't stop that from being implemented, since they are separate requirements. The flspot annotation could probably be implemented separately using this PR as a starting point in the future.

shamilovtim:
Fixed merge conflict d33c7bc

shamilovtim:
@imaNNeoFighT since this is a self contained feature that doesn't affect any other functionality. If you want to refactor it in the future and do it a different way, why not just merge it for now? I'm sure 90% of use cases would want to use it as-is and without having the trouble of complicated programming that some kind of manual placement would require. To place any widget anywhere on the chart would require the end user to make algorithms using all sorts of internal functions like `getTopOffset()` `getChartSize()` and then there would need to be some way to translate dp to canvas position. I believe it's impractical and does nothing for 99% of use cases, which is what open source functionality is for. Beating ourselves up for small edge cases that a minority of users would use is pointless. Here we take care of all of that complicated work and just let anyone annotate an axis. If someone wants more or to build a feature that changes dp to canvas position with a stack they are welcome to build that. But I suspect it still wouldn't deprecate the simplicity and use case of this feature and this feature would stay even if the other one were created.

imaNNeoFighT:
Dude, come down :))
I agreed with your previous message, but forgot to mention it,
It's under review and will be merged ASAP.
I know what you mean, and they are completely valid,
I have a concern to show a widget instead of touch tooltips, it's highly needed by users to have a very customizable tooltip (but we should talk about it in a separate thread).
Thanks for your explanation.
Cheers!

imaNNeoFighT:
Hi,
I couldn't run this branch, there are some compile errors.
image

Could you please take a look, and change the target branch to `dev`?
Thanks!

imaNNeoFighT:
Hello?

shamilovtim:
too many merge conflicts, i am going to need to recreate this branch

imaNNeoFighT:
I'm sorry to hear that,
I was really busy and it was my fault.

shamilovtim:
no worries I'll take care of it

PR #186, opened 2020-01-30
This adds support for persisting the touch indicator and tooltip after a touch event is complete

dazza5000:


imaNNeoFighT:
How we should clear the indicators if `persistIndicatorsAndTooltips` is true?

dazza5000:
We could make a clearIndicators method

imaNNeoFighT:
I think we need to clear indicators when the user tapped on nothing. 
What you think?

dazza5000:
I like the idea, do we have a callback / way to determine if the user
didn't tap on anything?

imaNNeoFighT:
We've handled something like this in the PieChartSample2,
please take a look.

##### Sample 2 ([Source Code](/example/lib/scatter_chart/samples/scatter_chart_sample2.dart))

shamilovtim:
FYI: in our use case we don't want them cleared at all. We want wherever the user tapped last to persist.

PR #208, opened 2020-02-25
Feature: full height touchline

shamilovtim:
This is an addition to the touchline. Setting the bool to `true` will stop it at the top of the chart instead of at the dot.

image


imaNNeoFighT:
Thanks

PR #216, opened 2020-03-03
Fix RangeAnnotations links in doc

isoppp:
Fixed links in documents.
I guess the syntax of these links were the opposite.


imaNNeoFighT:
Thanks!

Issue #214, opened 2020-03-01
I have set the clipToBorder to true, but it is still out of bounds


AndersIves:
left side is out of bounds
![](https://github.com/AndersIves/photos/blob/master/1.png?raw=true)
how can i fix it like this
![](https://github.com/AndersIves/photos/blob/master/2.png?raw=true)

> Hi,
> Please provide us a reproducible code to solve the problem.
> Thanks!

This is the class I defined
```

class SingleLineShadowChart extends StatefulWidget {
  final List<FlSpot> sports;
  final String yUnitString;

  final int days;

  SingleLineShadowChart(this.sports, this.yUnitString, this.days);

  @override
  _SingleLineShadowChartState createState() => _SingleLineShadowChartState();
}

class _SingleLineShadowChartState extends State<SingleLineShadowChart> {
  List<Color> gradientColors() {
    return [
      Theme.of(context).accentColor,
    ];
  }

  double minX = 0;
  double maxX = 0;
  double minY = 0;
  double maxY = 0;

  String getXTitleFunction(double value) {
    DateTime dateTime =
    DateTime.fromMillisecondsSinceEpoch(ConvertUtils.dse2mse(value));
    return "${dateTime.month}-${dateTime.day}";
  }

  String getYTitleFunction(double value) {
    return "${value.floor()} ${widget.yUnitString}";
  }

  @override
  void initState() {
    // TODO: implement initState
    super.initState();
  }

  @override
  Widget build(BuildContext context) {
    maxX = ConvertUtils.mse2dse(DateTime.now().millisecondsSinceEpoch);
    minX = ConvertUtils.mse2dse(DateTime.now().subtract(Duration(days: widget.days)).millisecondsSinceEpoch);
    maxY = widget.sports[0].y;
    minY = widget.sports[0].y;
    widget.sports.forEach((i) {
      if (i.y < minY) {
        minY = i.y;
      }
      if (i.y > maxY) {
        maxY = i.y;
      }
    });
    maxY = maxY + 0.1 * maxY;
    minY = minY - 0.1 * minY;
    return LineChart(
      LineChartData(
        clipToBorder: true,
        // 边框信息
        borderData: FlBorderData(
          show: true,
          border: Border.all(
            color: Theme.of(context).accentColor,
            width: 1,
          ),
        ),
        gridData: FlGridData(
          show: true,
          drawVerticalLine: true,

          verticalInterval: ((maxX - minX) / 4).floor().toDouble(),
[...]

PR #230, opened 2020-03-12
Fix clipToBorder in line chart

### LINKED ISSUE #228: Clip the chart within it's border
I want to present series which can have huge outliers , e.g. single value skyrocketing into thousands, where majority is in the -10 to 10 range. This makes the chart unreadable and impossibile to read important details. 

See the picture:
skyrocket

I'd like to be able to "zoom in" and clip series within borders. Currently, we can restrain Y axis but the line still goes out of bound (as well as the tooltips).
clipped

Ideally, there should also be an indicator where line goes out of bound stating the outlier value (the thousands). If possibile, the tooltips should also be restrained within chart borders.

onliner10:
Fixes #228 and #214

### DIFF
--- a/lib/src/chart/line_chart/line_chart_painter.dart
+++ b/lib/src/chart/line_chart/line_chart_painter.dart
@@ -76,8 +76,9 @@ class LineChartPainter extends AxisChartPainter<LineChartData>
     }
 
     if (data.clipToBorder) {
-      /// save layer to clip it to border after lines drew
       canvas.saveLayer(Rect.fromLTWH(0, -40, size.width + 40, size.height + 40), Paint());
+
+      clipToBorder(canvas, size);
     }
 
     super.paint(canvas, size);
@@ -109,9 +110,6 @@ class LineChartPainter extends AxisChartPainter<LineChartData>
     }
 
     if (data.clipToBorder) {
-      removeOutsideBorder(canvas, size);
-
-      /// restore layer to previous state (after clipping the chart)
       canvas.restore();
     }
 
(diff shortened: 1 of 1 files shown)


lib/src/chart/line_chart/line_chart_painter.dart
```diff
@@ -139,6 +137,20 @@ class LineChartPainter extends AxisChartPainter<LineChartData>
     }
   }
 
+  void clipToBorder(ui.Canvas canvas, ui.Size size,) {
```
imaNNeoFighT: Functions defined according to the usage of them in the paint() method,
please reorder its definition.

lib/src/chart/line_chart/line_chart_painter.dart
```diff
@@ -139,6 +137,20 @@ class LineChartPainter extends AxisChartPainter<LineChartData>
...
+    final Rect rect = Rect.fromLTRB(
+      getLeftOffsetDrawSize() - halfStrokeWidth,
+      getTopOffsetDrawSize() - halfStrokeWidth,
+      size.width -
+          (getExtraNeededHorizontalSpace() - getLeftOffsetDrawSize()) +
+          halfStrokeWidth,
+      size.height- (getExtraNeededVerticalSpace() - getTopOffsetDrawSize()) + halfStrokeWidth,
+    );
+
+    canvas.clipRect(rect, clipOp: ClipOp.intersect);
```
imaNNeoFighT: The default value of `clipOp` is `ClipOp.intersect`, then you don't have to to define it.

imaNNeoFighT:
And thank you, I didn't know that we have a `canvas.clipRect()` function.
<3
please change the target branch to `dev`

lib/src/chart/line_chart/line_chart_painter.dart
onliner10: Fixed

lib/src/chart/line_chart/line_chart_painter.dart
onliner10: I don't get it, it's the first function invoked within paint method and it's also declared right after paint

onliner10:
@imaNNeoFighT Rebased onto dev. Could you please look again?

lib/src/chart/line_chart/line_chart_painter.dart
imaNNeoFighT: You're right.

PR #231, opened 2020-03-12
Fix for fitInsideTheChart when tooltip is below or above the chart area

onliner10:
Fixes situation where the tooltip, having set fitInsideTheChart, goes above or below the chart area.

Was:
was
Is:
is

### DIFF
--- a/lib/src/chart/line_chart/line_chart_painter.dart
+++ b/lib/src/chart/line_chart/line_chart_painter.dart
@@ -973,6 +973,23 @@ class LineChartPainter extends AxisChartPainter<LineChartData>
             rect.bottom,
           );
       }
+
+      if (rect.top < 0) {
+        rect = Rect.fromLTRB(rect.left,
+          0,
+          rect.right,
+          tooltipHeight,
+        );
+      }
+
+      if (rect.bottom > viewSize.height) {
+        final shiftAmount = rect.bottom - viewSize.height + tooltipHeight;
+        rect = Rect.fromLTRB(rect.left,
+          rect.top - shiftAmount,
+          rect.right,
+          rect.bottom - shiftAmount,
+        );
+      }
     }
 
     final Radius radius = Radius.circular(tooltipData.tooltipRoundedRadius);


lib/src/chart/line_chart/line_chart_painter.dart
```diff
@@ -971,6 +971,24 @@ class LineChartPainter extends AxisChartPainter<LineChartData>
             rect.bottom,
           );
       }
+
+      if (rect.top < 0) {
+        final shiftAmount = rect.top;
```
imaNNeoFighT: I think we should have a positive value as a `shiftAmount`, we should shift it downward.
and shift the bottom like this:
`rect.bottom + shiftAmount`
it is simpler.

see line 958:
`final shiftAmount = 0 - rect.left;`

lib/src/chart/line_chart/line_chart_painter.dart
```diff
@@ -971,6 +971,24 @@ class LineChartPainter extends AxisChartPainter<LineChartData>
...
+        rect = Rect.fromLTRB(rect.left,
+          0,
+          rect.right,
+          rect.bottom - shiftAmount,
+        );
+      }
+
+      if (rect.bottom > viewSize.height) {
```
imaNNeoFighT: I think it is impossible to have a tooltip that passes the bottom border
because we show tooltips on top of points.
we can remove this section.

imaNNeoFighT:
And please change the target to `dev`.
Thanks!

lib/src/chart/line_chart/line_chart_painter.dart
onliner10: When you have multiple series going below the chart area it's possible

lib/src/chart/line_chart/line_chart_painter.dart
onliner10: Actually, I got rid of shiftAmount there entirely and imho it's even simpler now:
```
rect = Rect.fromLTRB(rect.left,
          0,
          rect.right,
          tooltipHeight,
        );
```

What do you think ?

onliner10:
@imaNNeoFighT I changed the base branch to dev, and squashed all commits into one so it's nice and clean.

lib/src/chart/line_chart/line_chart_painter.dart
onliner10: Actually, even something like this is enough to prove the point
image

imaNNeoFighT: Okay fair enough :)

lib/src/chart/line_chart/line_chart_painter.dart
imaNNeoFighT: It's great, 
Thanks!

PR #248, opened 2020-03-16
List interpolation

jamesblasco:
Before there was no animation between items in a list with different size.

For example in a line chart if a new FLSpot was added the graph was not animated. With this the spots that were before are animated if their position changed


imaNNeoFighT:
Thanks,
I was thinking of a better solution that handles adding or removing items animation, 
BTW, It's better than the current version,
I'll merge it, but think about what I've said.
Thanks for reporting.

imaNNeoFighT:
Could you please change the target branch to `dev`?

jamesblasco:
Same, I was thinking the same. I tried an algorithm that identifies the closest previous spot and animate from there but the performance was not very good.

imaNNeoFighT:
I think there are some list interpolations in the Flutter SDK, we should look at them.
Thanks!

PR #252, opened 2020-03-20
Feature: Nullable values in line chart

### LINKED ISSUE #26: Ability to show gaps in data?
Again, thanks for this plugin. You're doing an awesome job!

Is it possible to show gaps or null data? I've played around and haven't found how one might do that.

Visually, something along these lines:

image

Implementation wise, this might look like:

```
                    FlSpot(1, 2),
                    FlSpot(2, 3),
                    FlSpot(3, 3),
                    FlSpot(4, null),
                    FlSpot(5, null),
                    FlSpot(6, 2),
                    FlSpot(7, 4),
```

Thanks again!

### LINKED ISSUE #123: How to create an empty coordinate?
I need to use a list of coordinate to plot math function `ln(x)` where `x` ranges from `-5` to `5`, but it doesn't seem to have a way to create empty coordinates where `x` ranges from `-5` to `0`.

I tried to use `FlSpot(-2.0, null)` or `null` to represent an empty coordinate, but none of these work.

Are there any solutions here?

shamilovtim:
This PR provides the ability for a user to provide null points on the line chart to achieve gaps in data. 
`FlSpot(null,null)`

Sample has been updated. It's not pretty but it demonstrates a lot of different stuff.
image

Closes #26 
Closes #123


shamilovtim:
there are formatting changes again because of the last PR. contributor did not use dart formatter at all

lib/src/chart/line_chart/line_chart_painter.dart
```diff
@@ -166,19 +165,40 @@ class LineChartPainter extends AxisChartPainter<LineChartData>
...
-    _drawBelowBar(canvas, viewSize, belowBarPath, completelyFillAboveBarPath, barData);
-    _drawAboveBar(canvas, viewSize, aboveBarPath, completelyFillBelowBarPath, barData);
-    _drawBar(canvas, viewSize, barPath, barData);
+    // paint each sublist that was built above
+    // bar is passed in seperately from barData
+    // because barData is the whole line
+    // and bar is a piece of that line
+    barList.forEach((bar) {
+      final barPath = _generateBarPath(viewSize, barData, bar: bar);
+      _drawBar(canvas, viewSize, barPath, barData);
```
imaNNeoFighT: Why you changed the draw order?
previously _drawBar was the last one.

lib/src/chart/line_chart/line_chart_painter.dart
```diff
@@ -266,15 +294,16 @@ class LineChartPainter extends AxisChartPainter<LineChartData>
   /// and we use isCurved to find out how we should generate it,
   /// If you want to concatenate paths together for creating an area between
   /// multiple bars for example, you can pass the appendToPath
-  Path _generateBarPath(Size viewSize, LineChartBarData barData, {Path appendToPath}) {
+  Path _generateBarPath(Size viewSize, LineChartBarData barData,
+      {Path appendToPath, List<FlSpot> bar}) {
```
imaNNeoFighT: Why you added `bar` as an optional parameter?
I think it should not be optional.

lib/src/chart/line_chart/line_chart_painter.dart
```diff
@@ -339,13 +368,13 @@ class LineChartPainter extends AxisChartPainter<LineChartData>
   /// if [fillCompletely] is true, the cutOffY will be ignored,
   /// and a completely filled path will return,
   Path _generateBelowBarPath(Size viewSize, LineChartBarData barData, Path barPath,
-      {bool fillCompletely = false}) {
+      {List<FlSpot> bar, bool fillCompletely = false}) {
```
imaNNeoFighT: Also here `bar` is optional.

lib/src/chart/line_chart/line_chart_painter.dart
```diff
@@ -377,13 +406,13 @@ class LineChartPainter extends AxisChartPainter<LineChartData>
   /// if [fillCompletely] is true, the cutOffY will be ignored,
   /// and a completely filled path will return,
   Path _generateAboveBarPath(Size viewSize, LineChartBarData barData, Path barPath,
-      {bool fillCompletely = false}) {
+      {List<FlSpot> bar, bool fillCompletely = false}) {
```
imaNNeoFighT: And here `bar` is optional.

lib/src/chart/line_chart/line_chart_painter.dart
imaNNeoFighT: If it is for readability, please make it required with `@required` [annotation](https://dart.dev/guides/language/language-tour#optional-parameters).
And I think `bar` is not descriptive enough, I think `barSpots` is more descriptive.
Thanks!

imaNNeoFighT:
I'll add a linter check to verify it in each PR.
But please revert them, it makes confusion for now and later.

lib/src/chart/line_chart/line_chart_painter.dart
shamilovtim: Good catch

shamilovtim: Hmm 'bar' makes sense here because in the forEach it's iterating a barList so the two names connect. barSpotsList would be out of context

shamilovtim: I'll change the variable name to barSpots, I think that's what you were asking for

lib/src/chart/line_chart/line_chart_painter.dart
imaNNeoFighT: I'm asking for change this `bar` to `barSpots`.

imaNNeoFighT:
Thank you,
I'm going to the final review.
I think there was an issue for this feature,
could you please find it and mention it in the PR's title?

shamilovtim:
Just took care of the requests but I had another realization that I need to check sorting now that I'm making sublists

shamilovtim:
@imaNNeoFighT Unrelated to this PR but this is what happens on master right now when I provide

```dart
       spots: const [
            FlSpot(0, 1),
            FlSpot(11, 4),
            FlSpot(2, 1),
            FlSpot(4.9, 5),
            FlSpot(6.8, 5),
            FlSpot(8, 1),
            FlSpot(9.5, 2),
          ],
```

image


Should the spots be sorted or would that be incorrect for a line chart? It seems to be the `x` or `y` should determine the order, not the sequence of the elements in the array. Of course, we can always fix it in a different PR. I just wanted to call it out.

shamilovtim:
Ok all set. Just ignore my last comment, it's unrelated.

lib/src/chart/line_chart/line_chart_painter.dart
shamilovtim: Sorry let me fix

shamilovtim: done

lib/src/chart/line_chart/line_chart_painter.dart
shamilovtim: done

lib/src/chart/line_chart/line_chart_painter.dart
shamilovtim: done

lib/src/chart/line_chart/line_chart_painter.dart
shamilovtim: done

imaNNeoFighT:
About your last comment, I think it's correct :)
because you provided unordered spots, btw you can open an issue to discuss it there.
and about the PR, everything is ready to merge, but we should update the `LineChartBarData.spots` documentation, (dart docs, above the field and constructor), I will handle it,
Thanks!

imaNNeoFighT:
I've made some changes on your codes, they are for simplification check the commit [here](https://github.com/imaNNeoFighT/fl_chart/commit/555c473205c4c65904b1129162c7e892a581b4b3).

jacobokoenig:
Hi, was this ever merged to the main branch, or is there an alternative to create empty or invisible sections in a line chart when x is null? I have a use case where data for a specific x point is not there, but I can't consider it zero, nor can I draw a line to the next point and assume the data just changed linearly.

Thanks!

shamilovtim:
@jacobokoenig Yeah I implemented this and it was merged many years ago and used in production. Unsure of its current status.

jacobokoenig:
@shamilovtim doesn't look like we can include null values. I found a workaround using multiple line segments by dividing my data into arrays (although not the ideal situation). Thanks!

PR #262, opened 2020-03-25
Fix #261

### LINKED ISSUE #261: Tapping a dot doesn't work when there are multiple dots with the same X value
I found another bug. When you have multiple values with the same X value but different Y values, the detection mechanism does not work, you can select only one dot.
Example:
example
In this case only 4.10% can be chosen.

tudor07:
Fixed issue: https://github.com/imaNNeoFighT/fl_chart/issues/261


imaNNeoFighT:
Too many commits :))

Issue #210, opened 2020-02-28
how to handle normal taps on a pie section with touchCallback 


Sarathlal13:
FlLongPressStart, FlLongPressMoveUpdate, FlLongPressEnd, FlPanStart, FlPanMoveUpdate, FlPanEnd with these touch behaviors how can i trigger a normal button tap on a pie section. i want to navigate to  a new page on tap

imaNNeoFighT:
Hi, great question, unfortunately, we didn't support it built-in (due to some complexity).
but there is a sample that we've implemented onTap behavior,
check it out:
##### Sample 2 ([Source Code](/example/lib/scatter_chart/samples/scatter_chart_sample2.dart))
<img src="https://github.com/imaNNeoFighT/fl_chart/raw/master/repo_files/images/scatter_chart/scatter_chart_sample_2.gif" width="300" >

imaNNeoFighT:
But let it be open, we will handle it built-in ASAP.
Stay tuned, thank yout!

Sarathlal13:
thanks bro!

imaNNeoFighT:
You're welcome ;)

imaNNeoFighT:
We have just added `clickHappened` property in the touch responses.
check it out in [0.30.0](https://github.com/imaNNeoFighT/fl_chart/releases/tag/0.30.0)

We used this property in:

##### Scatter Sample 2 ([Source Code](https://github.com/imaNNeoFighT/fl_chart/blob/master/example/lib/scatter_chart/samples/scatter_chart_sample2.dart))
<img src="https://github.com/imaNNeoFighT/fl_chart/raw/master/repo_files/images/scatter_chart/scatter_chart_sample_2.gif" width="300" >

Issue #202, opened 2020-02-15
LineChart inside PageView breaks the pan gesture.


jamesblasco:
In the example project, `LineChart` doesn't change the tooltip when dragging  through the widget.
This is because `PageView` is using a `HorizontalDragGestureRecognizer` that has priority to the `PanGestureRecognizer` that is used on the `LineChart`(the gesture detector widget with onPanStart...)

My proposal is to add:

onHorizontalDragDown 
onHorizontalDragStart 
onHorizontalDragUpdate
onHorizontalDragEnd 
onHorizontalDragCancel

with the same params of

onPanDown
onPanStart
onPanUpdate
onPanEnd
onPanCancel

This will make the LineChart have priority to the PageView.

To make it more adaptable we could add the param `draggable:bool` to `LineTouchData` in case people want to disable the pan gesture to prioritize other gestures in the Gesture Arena

imaNNeoFighT:
Hi,
I didn't think about it, because I intentionally restricted it to work with long-press if it is in a scrollable parent.
but let me think about it, and let it be open.
Thanks for reporting.

jamesblasco:
Thanks! I understand why you would want it like that. I will use my fork in the meantime.

Amazing work you are doing!! Thanks again

dkobia:
I too I'm experiencing an issue with a barchart inside a SingleChildScrollview:

```
The following NoSuchMethodError was thrown while handling a gesture:
The getter 'length' was called on null.
Receiver: null
Tried calling: length

When the exception was thrown, this was the stack: 
#0      Object.noSuchMethod (dart:core-patch/object_patch.dart:53:5)
#1      BarChartPainter._getNearestTouchedSpot (package:fl_chart/src/chart/bar_chart/bar_chart_painter.dart:577:43)
#2      BarChartPainter.handleTouch (package:fl_chart/src/chart/bar_chart/bar_chart_painter.dart:564:9)
#3      _BarChartState.build.<anonymous closure> (package:fl_chart/src/chart/bar_chart/bar_chart.dart:84:58)
#4      GestureRecognizer.invokeCallback (package:flutter/src/gestures/recognizer.dart:182:24)
...
Handler: "onCancel"
Recognizer: PanGestureRecognizer#defd2
  debugOwner: GestureDetector
  start behavior: start
```

[...]

Issue #73, opened 2019-09-28
Is it possible to set the color of bar when we touch on it?


nimesh1997:
When I touch on the bar chart I want to set the colour on that particular bar and should not disappear when I remove my hand?

imaNNeoFighT:
Check it out:

![bar_click](https://user-images.githubusercontent.com/7009300/111708641-761dd480-885b-11eb-860b-d1aacecd885c.gif)

[SourceCode](https://gist.github.com/imaNNeoFighT/7eaf2ac51c9b870e5c61abd47275a328)

PR #264, opened 2020-03-25
Sticky touch indicator

shamilovtim:
This time I have placed the boolean on the lineTouchData along with the rest of the touch configuration.  Very simple but useful. Please consider merging.

PR #268, opened 2020-03-26
Replace pan gesture with horizontal drag & tap combination

### LINKED ISSUE #227: Scroll gestures in an Android scrollview trigger a touchline
As seen in the gif. The touchline is vertical, which means only horizontal swipes should trigger it. However swiping up and down on the scrollview triggers the touchline. This is not ideal behavior.

image

### LINKED ISSUE #265: Bug: panEnd gesture does not get the right location on the line always gets 1st point
The panEnd gesture always returns the 1st point on the barline, it does not return the correct spot it was stopped. I was able to view this by logging out where the panEnd stops. It should stop at the last update point but it resets back to one.

image

shamilovtim:
The previous "Pan" gesture was causing problems on scroll views where a quick swipe up would trigger a touch response inside of a chart. It has been replaced with a combination of horizontal drag and tap. The tap is only there so that the first swiped point is shown, which triggers a tapCancel. It's a little complicated but it works correctly to replace pan. I reused some of the old pan constructors for simplicity and backward compatibility. 

Furthermore, I used setState to get the localPosition from the gesture update in order to properly provide the final updated position to the end of the gesture. 

Fixes #265, #227 , #210, #202. 
Also if we pass the tapped line in the LineChart, this #73, gonna be fixed. 
It should be checked #276.

### DIFF
--- a/lib/src/chart/line_chart/line_chart.dart
+++ b/lib/src/chart/line_chart/line_chart.dart
@@ -9,7 +9,6 @@ import 'line_chart_painter.dart';
 
 /// Renders a line chart as a widget, using provided [LineChartData].
 class LineChart extends ImplicitlyAnimatedWidget {
-
   /// Determines how the [LineChart] should be look like.
   final LineChartData data;
 
@@ -24,7 +23,6 @@ class LineChart extends ImplicitlyAnimatedWidget {
   /// Creates a [_LineChartState]
   @override
   _LineChartState createState() => _LineChartState();
-
 }
 
 class _LineChartState extends AnimatedWidgetBaseState<LineChart> {
@@ -39,6 +37,9 @@ class _LineChartState extends AnimatedWidgetBaseState<LineChart> {
   final List<MapEntry<int, List<LineBarSpot>>> _showingTouchedTooltips = [];
 
   final Map<int, List<int>> _showingTouchedIndicators = {};
+
+  // to set the correct ending position of a gesture
+  Offset _lastTouchedPosition;
 
   @override
   Widget build(BuildContext context) {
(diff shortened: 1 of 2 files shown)


lib/src/chart/line_chart/line_chart.dart
```diff
@@ -82,31 +85,46 @@ class _LineChartState extends AnimatedWidgetBaseState<LineChart> {
...
         final Size chartSize = _getChartSize();
         if (chartSize == null) {
           return;
         }
 
         final LineTouchResponse response =
-            _touchHandler?.handleTouch(FlPanEnd(Offset.zero, details.velocity), chartSize);
+            _touchHandler?.handleTouch(FlPanStart(details.localPosition), chartSize);
         if (_canHandleTouch(response, touchData)) {
+          setState(() {
```
imaNNeoFighT: Why do you call `setState()` here?

lib/src/chart/line_chart/line_chart.dart
```diff
@@ -82,31 +85,46 @@ class _LineChartState extends AnimatedWidgetBaseState<LineChart> {
...
+          return;
+        }
+
+        final LineTouchResponse response =
+            _touchHandler?.handleTouch(FlTap(details.localPosition), chartSize);
+        if (_canHandleTouch(response, touchData)) {
+          touchData.touchCallback(response);
+        }
+      },
+      onHorizontalDragStart: (DragStartDetails details) {
```
imaNNeoFighT: Do you think we just need horizontal?
do we need vertical direction too?
I think it's needed when we have a vertical scroll view.

shamilovtim: I don't see how we could need a vertical gesture because we only have a vertical touchline, which respond to a horizontal gesture. If we had a horizontal touchline, it would be a different story. But linecharts are left to right, never up and down, so I don't see how we would ever need it.

lib/src/chart/line_chart/line_chart.dart
shamilovtim: Because `onTapCancel` needs that position. Otherwise you would have the problem of needing to pass a fake `Offset.zero` which is not actually where the tap got cancelled.

imaNNeoFighT: No no,
I mean ` _lastTouchedPosition = details.localPosition;` is enough,
why did you put `setState()` around it?
`setState()` leads to call build mthod again.

lib/src/chart/line_chart/line_chart.dart
imaNNeoFighT: We have to try a solution that works on all type of charts
and even in the LineChart, maybe you have some spots in the same x-axis and different y values.

lib/src/chart/line_chart/line_chart.dart
shamilovtim: Because it's a stateful variable from the previous update. Maybe I'm confused

lib/src/chart/line_chart/line_chart.dart
shamilovtim: The long press still exists for those points. A vertical drag would, in theory, only be needed if he had a horizontal touchline to start swiping up and down.

lib/src/chart/line_chart/line_chart.dart
shamilovtim: Just tested, you're right

shamilovtim:
Just tested vertical dragging under this implementation and it seems right to me. A swipe up on the chart scrolls the view. A horizontal drag lets you swipe up and down, and a long press also lets you swipe up and down. 
Screen Recording 2020-03-26 at 12 43 54 PM

lib/src/chart/line_chart/line_chart.dart
shamilovtim: See my comment below if that's acceptable

lib/src/chart/line_chart/line_chart.dart
shamilovtim: Removed setstate

lib/src/chart/line_chart/line_chart.dart
shamilovtim: Also: this solution is only for the line chart. The other charts use separate gestures in separate files. 

Regarding multiple different y values. Even if a user has multiple different values, he doesn't want his chart to stop his scrollview from scrolling. That's why we block this gesture, it conflicts with standard mobile UX behavior. They still have access to Y points up and down by starting with the correct gesture long press, tap or right swipe. Just not upswipe.

imaNNeoFighT:
I think it's the most useful feature in FLChart's lifetime :)
I mentioned issues that gonna be closed if this merged

shamilovtim:
Good to hear

imaNNeoFighT:
Hi @shamilovtim,
It seems you've used sticky touch #264 feature, in this code.
Then I have to first merge that.

shamilovtim:
Branched off the wrong branch I guess. Pretty simple to rebase out

shamilovtim:
I was having some problems in my fork so I think that's what this came from. I should be able to reset it

imaNNeoFighT:
Okay, I'm waiting for you.
Thanks!

shamilovtim:
all set but the linebreak troubles have appeared again

imaNNeoFighT:
Hi @shamilovtim,
You've closed sticky touch indicator branch, and I remember you used it in this branch,
what should we do with this branch now?

imaNNeoFighT:
I'm going to bump the version into **0.9.0**, and it's gonna be great if we ship this feature.
let me know what do you think?

shamilovtim:
The branch has been ready and done since my last comment. Sticky I couldn’t get in because it would have a merge conflict almost every day so I gave up on it. It’s not part of this branch anymore I rebased it out a while ago.

imaNNeoFighT:
Ok thanks!

PR #279, opened 2020-04-01
Update roadmap.md

primemb:
add Range Column Chart like this : https://apexcharts.com/react-chart-demos/column-charts/range-column-chart/


imaNNeoFighT:
Hi,
we've found that our roadmap isn't efficient enough,
because we should take our steps based on people's demands.
please mane an issue for it and drop that sample as an image,
and people can vote it up using thumbs-up emoji.
Thanks for understanding!

PR #289, opened 2020-04-06
Fix code blocks highlighting

mhmdanas:
The code blocks are now highlighted correctly, whereas before there was no syntax highlighting or wrong syntax highlighting.


imaNNeoFighT:
Thanks!

PR #301, opened 2020-04-16
Adding barData parameter to checkToShowDot Function

krispypen:
Like the GetDotColorCallback that contains the barData parameter, it's usefull to pass it also to the checkToShowDot function. In my case I needed to know the index position of the spot for example. That was not possible to find out while lerp-ing (animation).


imaNNeoFighT:
Okay please change the target branch to `dev`

krispypen:
Now it also contains some stuff that is only done in the master branch. I think that should also be merged? If not I will recreate my branch starting from dev and add only my commit

imaNNeoFighT:
It's gonna be in the next version (about one hour)
Thanks!

PR #305, opened 2020-04-19
Feature/line chart shadow

imaNNeoFighT:


krispypen:
If I just upgrade to 0.9.2 we alsways get shadows? looks like a bug? I had a line with color transparent, and now it's showing... until i added a shadow property Shadow(color: Colors.transparent) Or is this expected behavior?

imaNNeoFighT:
No, this is not expected behavior, please file an issue.
Thanks!

PR #319, opened 2020-04-23
Shadows default off

### LINKED ISSUE #316: Black shadows are default on since 0.9.2
Since version 0.9.2 , PR #305 , every line has a shadow, this should not be standard behavior.

My suggestion is to add a color property to the default " ?? const Shadow()" at https://github.com/imaNNeoFighT/fl_chart/blob/ef6ee3f882cbb01a5661c998fa67633f1c23144d/lib/src/chart/line_chart/line_chart_data.dart#L393
and set it to Colors.transparent . Then it's ignored during rendering in the _drawBarShadow method because the opacity is 0.0

Not 100% sure but if that's the best solution, I can create a PR too

krispypen:
Fixes #316


imaNNeoFighT:
Please change the target branch to dev.
Thanks!

imaNNeoFighT:
And did you test it?
Is it working?

imaNNeoFighT:
Thanks!

PR #339, opened 2020-05-08
Update line_chart_painter.dart

AmarjotSingh21:
A range error is thrown in _generateBarPath() if we add consecutive null values
for eg: [ FlSpot(1,2), FlSpot(null, null), FlSpot(null, null), FlSpot(4, 4)] will throw range error
(i know its stupid to add consecutive null values but i am that stupid i couldn't understand why m getting range error)

I added  a condition which will take only first null value if there are consecutive null values


imaNNeoFighT:
Hi, there was a problem with `master`, that's why this branch is failed.
Thanks.

PR #348, opened 2020-05-22
Added option to change shape of line chart dots

jlubeck:
I was looking for a way to change the shape of the dots in the same line depending on a value and I couldn't see support for it. So I went ahead and made it myself.

Hopefully you find it useful and can merge it.

It could be easily expanded to use other shapes. 

Let me know if you think I should change anything.

### DIFF
--- a/example/lib/line_chart/samples/line_chart_sample1.dart
+++ b/example/lib/line_chart/samples/line_chart_sample1.dart
@@ -395,15 +395,7 @@ class LineChartSample1State extends State<LineChartSample1> {
         ],
         barWidth: 2,
         isStrokeCapRound: true,
-        dotData: FlDotData(
-            show: true,
-            getDotDrawer: (spot, percentage, data, index) {
-              final FlDotDrawer circle = FlDotCircleDrawer(
-                  radius: 10, color: Colors.red, strokeWidth: 3, strokeColor: Colors.white);
-              final FlDotDrawer square = FlDotSquareDrawer(
-                  size: 20, color: Colors.green, strokeWidth: 2, strokeColor: Colors.orange);
-              return index % 2 == 0 ? circle : square;
-            }),
+        dotData: FlDotData(show: true),
         belowBarData: BarAreaData(
           show: false,
         ),
--- a/example/lib/line_chart/samples/line_chart_sample1.dart
+++ b/example/lib/line_chart/samples/line_chart_sample1.dart
@@ -396,8 +396,15 @@ class LineChartSample1State extends State<LineChartSample1> {
         barWidth: 2,
         isStrokeCapRound: true,
         dotData: FlDotData(
-          show: true,
-        ),
+            show: true,
+            dotSize: 20,
+            getDotDrawer: (spot, percentage, data, index) {
+              final FlDotDrawer circle =
+                  FlDotCircleDrawer(color: Colors.red, strokeWidth: 3, strokeColor: Colors.white);
+              final FlDotDrawer square = FlDotSquareDrawer(
+                  size: 10, color: Colors.green, strokeWidth: 2, strokeColor: Colors.orange);
+              return index % 2 == 0 ? circle : square;
+            }),
         belowBarData: BarAreaData(
           show: false,
         ),
--- a/example/lib/line_chart/samples/line_chart_sample1.dart
+++ b/example/lib/line_chart/samples/line_chart_sample1.dart
@@ -397,12 +397,11 @@ class LineChartSample1State extends State<LineChartSample1> {
         isStrokeCapRound: true,
         dotData: FlDotData(
             show: true,
-            dotSize: 20,
             getDotDrawer: (spot, percentage, data, index) {
-              final FlDotDrawer circle =
-                  FlDotCircleDrawer(color: Colors.red, strokeWidth: 3, strokeColor: Colors.white);
+              final FlDotDrawer circle = FlDotCircleDrawer(
+                  radius: 10, color: Colors.red, strokeWidth: 3, strokeColor: Colors.white);
               final FlDotDrawer square = FlDotSquareDrawer(
-                  size: 10, color: Colors.green, strokeWidth: 2, strokeColor: Colors.orange);
+                  size: 20, color: Colors.green, strokeWidth: 2, strokeColor: Colors.orange);
               return index % 2 == 0 ? circle : square;
             }),
         belowBarData: BarAreaData(
--- a/example/lib/line_chart/samples/line_chart_sample8.dart
+++ b/example/lib/line_chart/samples/line_chart_sample8.dart
@@ -154,8 +154,8 @@ class _LineChartSample8State extends State<LineChartSample8> {
             return TouchedSpotIndicatorData(
               FlLine(color: Colors.orange, strokeWidth: 3),
               FlDotData(
-                  getDotDrawer: (spot, percent, barData, index) =>
-                      FlDotCircleDrawer(color: Colors.deepOrange)),
+                  getDotPainter: (spot, percent, barData, index) =>
+                      FlDotCirclePainter(radius: 8, color: Colors.deepOrange)),
             );
           }).toList();
         },
--- a/example/lib/line_chart/samples/line_chart_sample8.dart
+++ b/example/lib/line_chart/samples/line_chart_sample8.dart
@@ -153,7 +153,9 @@ class _LineChartSample8State extends State<LineChartSample8> {
           return spotIndexes.map((spotIndex) {
             return TouchedSpotIndicatorData(
               FlLine(color: Colors.orange, strokeWidth: 3),
-              FlDotData(dotSize: 8, getDotColor: (spot, percent, barData) => Colors.deepOrange),
+              FlDotData(
+                  getDotDrawer: (spot, percent, barData, index) =>
+                      FlDotCircleDrawer(color: Colors.deepOrange)),
             );
           }).toList();
         },
--- a/lib/src/chart/line_chart/line_chart_data.dart
+++ b/lib/src/chart/line_chart/line_chart_data.dart
@@ -784,29 +784,15 @@ Color _defaultGetDotStrokeColor(FlSpot spot, double xPercentage, LineChartBarDat
 /// It should return a [FlDotDrawer] that needs to be used for drawing target.
 typedef GetDotDrawerCallback = FlDotDrawer Function(FlSpot, double, LineChartBarData, int);
 
-FlDotDrawer _defaultGetDotDrawer(FlSpot spot, double xPercentage, LineChartBarData bar, int index) {
-  return FlDotCircleDrawer(
-      size: bar.dotData.dotSize, color: _defaultGetDotColor(spot, xPercentage, bar));
+FlDotDrawer _defaultGetDotDrawer(FlSpot spot, double xPercentage, LineChartBarData bar, int index,
+    {double size}) {
+  return FlDotCircleDrawer(radius: size, color: _defaultGetDotColor(spot, xPercentage, bar));
 }
 
 /// This class holds data about drawing spot dots on the drawing bar line.
 class FlDotData with EquatableMixin {
   /// Determines show or hide all dots.
   final bool show;
-
-  // Customizes the size of the dot
-  final double dotSize;
-
-  /// The stroke width to use for the corresponding [FlSpot]
-  final double strokeWidth;
-
-  /// Callback which is called to the the stroke color of the given [FlSpot].
-  /// The [FlSpot] is provided as parameter to this callback
-  final GetDotColorCallback getStrokeColor;
-
-  /// Callback which is called to the the color of the given [FlSpot].
-  /// The [FlSpot] is provided as parameter to this callback
-  final GetDotColorCallback getDotColor;
 
   /// Checks to show or hide an individual dot.
   final CheckToShowDot checkToShowDot;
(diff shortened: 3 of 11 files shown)


imaNNeoFighT:
Hi, there was a problem with `master`, that's why this branch is failed.
Thanks.

jlubeck:
Hey! I just updated the branch. Looks like the test passed but not the code style. But when I try to look at the details, it just says `Formatting failed: 1 ##[error]Process completed with exit code 1.`

If you can tell me what the problem is I'll fix it!

imaNNeoFighT:
Okay, thanks!
As described in our [contributions](https://github.com/imaNNeoFighT/fl_chart/blob/master/CONTRIBUTING.md) guideline, you should format the code before committing by `flutter format --line-length 100 .` command.
Let me know if it doesn't make sense.
Thanks!

jlubeck:
Perfect sense, thank you. Don't know why VSCode didn't do it automatically for me, but running the formatter manually fixed it. Thanks!

imaNNeoFighT:
Well done, I'm gonna review it.

lib/src/chart/line_chart/line_chart_data.dart
```diff
@@ -846,9 +866,15 @@ class FlDotData with EquatableMixin {
         getStrokeColor,
         getDotColor,
         checkToShowDot,
+        getDotShape,
       ];
 }
 
+enum FlDotDataShape {
```
imaNNeoFighT, line 873: I think enum will restrict us in the future,
because we are trying to make it possible to provide a widget to show,
but with the enum, it's impossible or overkill,
On the other hand, maybe we want to customize the shape, for example, if we decide to have a rounded radius in a square shape enum doesn't let us do that.
My suggestion is to have an abstractClass paradigm like this:

```dart
abstract class FlDotDrawer {
  void draw(Canvas canvas, FlSpot spot, Offset offsetInCanvas);
}

class FlDotCircleDrawer extends FlDotDrawer {

  double size;
  double strokeWidth;
  Color color;

  @override
  void draw(Canvas canvas, FlSpot spot, Offset offsetInCanvas) {
    canvas.drawCircle(offsetInCanvas, size, Paint()..color = color);
  }

}

class FlDotSquareDrawer extends FlDotDrawer {

  Color color;
  double size;
  Radius radius;

  @override
  void draw(Canvas canvas, FlSpot spot, Offset offsetInCanvas) {
    final halfSize = size / 2;
    final left = offsetInCanvas.dx - halfSize;
    final top = offsetInCanvas.dy - halfSize;
    final right = offsetInCanvas.dx + halfSize;
    final bottom = offsetInCanvas.dy + halfSize;
    canvas.drawRRect(RRect.fromLTRBR(left, top, right, bottom, radius), Paint()..color = color);
  }

}
```

then we can have a reference to the `FlDotDrawer` class, and use `FlDotDrawer.draw` to draw it in the line_chart_painter class,
it has lot's of benefits, like:

1. we are not restricted, and we can have any kind of shape that we want, like having cross shape, or anything that we can draw on the canvas.
 
2. each shape properties will be held inside of that class, instead of `FlDotData` (I mean all color, size,... properties should be removed from `FlDotData`, and each concrete class should hold it's own properties) 

3. each paint logic will be handled in the child concrete classes, and our painter class will stay as clear as possible.

Let me know what do you think.

Thanks!

imaNNeoFighT: I know it is a huge and time-consuming job, but I promise you, it will help people a lot.
and It's a professional way to do that.

It will help us to have any kind of shape that you think.
Thanks in advance!

imaNNeoFighT: And for sure, it's kind of pseudocode, I mean I'm not sure if is it working or not, I didn't run it.

jlubeck: Oh I totally agree. I was actually hoping when I first had the requirement that you were already using a Widget to draw the Dots and it would just be a matter of exposing that widget haha. After I saw the Canvas drawing I went for the enum just because I had an immediate problem to solve for our product, but I totally agree about this. I have some other things to finish first, but I'll try to tackle this asap. Thanks for the pointers and glad to hear you backed the idea

imaNNeoFighT: Yes, drawing the widget is the next step (It's kinda hard to do that, but we should think about it),
but for now, let's expose the canvas to draw anything we want (or anything end-user wants).
I'm going to bump up the version to 0.10.0,
It will be great if we introduce it in this release,
when do you think you can have it finished?

jlubeck: I will make my best to have it done today, if not, tomorrow

imaNNeoFighT: Thanks!

jlubeck:
So, I was able to implement the abstract class and it is working great. But, I started removing the other properties out of the `FlDotData` and ran into some issues because some of those properties (mainly size) are being used in a LOT of other places, and I couldn't figure out exactly how to switch it out. If you can give me some pointers there, I would gladly continue removing them from `FlDotData`

imaNNeoFighT:
Hi, I've checked your latest commit and it's looking great, good job man, keep rocking.
I've also checked the `size` property usages, we used it everywhere for showing the tooltip above it,
I mean for showing the tooltip we had to know the height of the dot to show it properly.
My suggestion is to have a 
```dart
Size getSize(FlSpot spot);
```
abstract function in our `FlDotDrawer` (and it should be implemented by the child classes), then call `dotDrawer.getSize().height` instead of `flDotData.dotSize` everywhere. (in previous approach, `dotSize` was our both width and height, because shape was circle, now we can have both width, and height by using [Size](https://api.flutter.dev/flutter/dart-ui/Size-class.html) class).

Let me know whether it is working or not.
Thanks!

jlubeck:
Ok, let's see what you think!

jlubeck:
Finished fixing the tests @imaNNeoFighT

imaNNeoFighT:
Are you done?

jlubeck:
Should be yes, unless you have more recommendations

imaNNeoFighT:
Okay, let me check it out,
BTW, Thanks you.

jlubeck:
Oh no, thank YOU.

lib/src/chart/line_chart/line_chart_data.dart
```diff
@@ -762,90 +761,159 @@ Color _defaultGetDotColor(FlSpot _, double xPercentage, LineChartBarData bar) {
...
+class FlDotCircleDrawer extends FlDotDrawer {
+  Color color;
+  double radius;
+  Color strokeColor;
+  double strokeWidth;
+
+  FlDotCircleDrawer({
+    @required this.color,
+    this.radius = 4.0,
+    this.strokeColor,
```
imaNNeoFighT, line 833: Please provide default `strokeColor` and `strokeWidth`, or just mark them as required (I prefer the first default values)

lib/src/chart/line_chart/line_chart_data.dart
```diff
@@ -762,90 +761,159 @@ Color _defaultGetDotColor(FlSpot _, double xPercentage, LineChartBarData bar) {
...
+class FlDotSquareDrawer extends FlDotDrawer {
+  Color color;
+  double size;
+  Color strokeColor;
+  double strokeWidth;
+
+  FlDotSquareDrawer({
+    @required this.color,
+    this.size = 4.0,
+    this.strokeColor,
```
imaNNeoFighT, line 879: And apply default values here please

lib/src/chart/line_chart/line_chart_data.dart
jlubeck: Well, if you don't provide a strokeWith it doesn't even draw the stroke. 
```
if (strokeWidth != null) {
      canvas.drawCircle(...)
}
```

And if you don't provide a strokeColor it just uses the color. 
```
Paint()..color = strokeColor ?? color
```

jlubeck: Isn't it more performant to just not draw it, instead of drawing a shape with 0 stroke?

imaNNeoFighT: Hmm, you're right, but I'm thinking in another direction,
If we put a thin stroke around the spots, people can realize that there is a way to have a stroke (they will find it sooner), and If they don't want it, they can find the `strokeWidth` property and set it to zero, I think it's better if you think about it's UX.
(previously we had a `_defaultGetDotStrokeColor`, you can revert it back and use it, and I think 1 stroke with is enough).
Thanks!

lib/src/chart/line_chart/line_chart_data.dart
```diff
@@ -762,90 +761,159 @@ Color _defaultGetDotColor(FlSpot _, double xPercentage, LineChartBarData bar) {
...
+
+class FlDotCircleDrawer extends FlDotDrawer {
+  Color color;
+  double radius;
+  Color strokeColor;
+  double strokeWidth;
+
+  FlDotCircleDrawer({
+    @required this.color,
+    this.radius = 4.0,
```
imaNNeoFighT, line 832: Please stick to our initializing constructor format for all constructor parameters,
like all other model constructors, it should be something like:

```dart
FlDotCircleDrawer({
    double radius,
  }) : radius = radius ?? 4.0;
```

lib/src/chart/line_chart/line_chart_data.dart
```diff
@@ -762,90 +761,159 @@ Color _defaultGetDotColor(FlSpot _, double xPercentage, LineChartBarData bar) {
...
-        dotSize,
-        strokeWidth,
-        getStrokeColor,
-        getDotColor,
         checkToShowDot,
+        getDotDrawer,
+      ];
+}
+
+abstract class FlDotDrawer with EquatableMixin {
```
imaNNeoFighT, line 818: Hmm, let's talk about "FlDot**Drawer**" name,
Do you think it's descriptive and unique?
I think people will be confused about it, because we have a [Drawer](https://api.flutter.dev/flutter/material/Drawer-class.html) widget in the Flutter.
and on the other hand, we have CustomPainter that its function is about draw something.
In my opinion, **FlDotPainter** is more descriptive and better.
What do you think?

example/lib/line_chart/samples/line_chart_sample1.dart
```diff
@@ -396,8 +396,14 @@ class LineChartSample1State extends State<LineChartSample1> {
         barWidth: 2,
         isStrokeCapRound: true,
         dotData: FlDotData(
-          show: true,
-        ),
+            show: true,
+            getDotDrawer: (spot, percentage, data, index) {
```
imaNNeoFighT, line 400: And about here, our feature shouldn't break anything, I mean it should work as same before (because we've added just a new type, and previous one is still supported).
I mean just by setting `FlDotData.show = true`, it should work like the previous version.

example/lib/line_chart/samples/line_chart_sample8.dart
```diff
@@ -153,7 +153,9 @@ class _LineChartSample8State extends State<LineChartSample8> {
           return spotIndexes.map((spotIndex) {
             return TouchedSpotIndicatorData(
               FlLine(color: Colors.orange, strokeWidth: 3),
-              FlDotData(dotSize: 8, getDotColor: (spot, percent, barData) => Colors.deepOrange),
+              FlDotData(
+                  getDotDrawer: (spot, percent, barData, index) =>
+                      FlDotCircleDrawer(color: Colors.deepOrange)),
```
imaNNeoFighT, line 158: You've forgotten to apply `dotSize: 8` in your changes.

imaNNeoFighT:
And I think after resolving these issues, It will be ready to merge,
Sorry about this long process, we should do our best for having branches merged and that's why code quality is important, by these common changes, we can maintain it better in the future.
Thanks in advance!

imaNNeoFighT:
Hi man,
any answer?

jlubeck:
Hey sorry, I had to go back to a different project. Looking at this now

lib/src/chart/line_chart/line_chart_data.dart
jlubeck: I agree. I was actually thinking about that other Drawer myself

jlubeck:
Ok @imaNNeoFighT , ready for a new review!

lib/src/chart/line_chart/line_chart_data.dart
```diff
@@ -776,76 +776,169 @@ Color _defaultGetDotStrokeColor(FlSpot spot, double xPercentage, LineChartBarDat
...
+  double strokeWidth;
+
+  FlDotCirclePainter({
+    Color color,
+    double radius,
+    Color strokeColor,
+    double strokeWidth,
+  })  : color = color ?? Colors.green,
+        radius = radius ?? 4.0,
+        strokeColor = strokeColor ?? Colors.green,
```
imaNNeoFighT, line 856: please make a differentiate between `color`, and `strokeColor`, I prefer `Colors.green.darken()` for strokeColor.

lib/src/chart/line_chart/line_chart_data.dart
```diff
@@ -776,76 +776,169 @@ Color _defaultGetDotStrokeColor(FlSpot spot, double xPercentage, LineChartBarDat
...
+  double strokeWidth;
+
+  FlDotSquarePainter({
+    Color color,
+    double size,
+    Color strokeColor,
+    double strokeWidth,
+  })  : color = color ?? Colors.green,
+        size = size ?? 4.0,
+        strokeColor = strokeColor ?? Colors.green,
```
imaNNeoFighT, line 905: Ane here please!

lib/src/chart/line_chart/line_chart_data.dart
```diff
@@ -776,76 +776,169 @@ Color _defaultGetDotStrokeColor(FlSpot spot, double xPercentage, LineChartBarDat
...
+}
+
+abstract class FlDotPainter with EquatableMixin {
+  void draw(Canvas canvas, FlSpot spot, Offset offsetInCanvas);
+
+  Size getSize(FlSpot spot);
+}
+
+class FlDotCirclePainter extends FlDotPainter {
+  Color color;
```
imaNNeoFighT, line 844: And the last one,
Please write a minimum document for these parameters and the constructor and the class,
because for some reason we need to increase our documentation coverage.
Thanks!

imaNNeoFighT:
And please update the line_chart.md.
I think after that, we are done!

jlubeck:
There you go @imaNNeoFighT!

imaNNeoFighT:
Congratulations!
You did it amazing.
Thanks a lot.

jlubeck:
Nice!! Thanks!

jlubeck:
Any idea when you will be merging dev to master?

imaNNeoFighT:
Soon!

PR #350, opened 2020-05-24
Removed Range Error if we add consecutive null values in line chart

AmarjotSingh21:
A range error is thrown in _generateBarPath() if we add consecutive null values
for eg: [ FlSpot(1,2), FlSpot(null, null), FlSpot(null, null), FlSpot(4, 4)] will throw range error


imaNNeoFighT:
Hi, there was a problem with master, that's why this branch is failed.
Thanks.

PR #361, opened 2020-05-29
Fix touch detection of sections spanned across the 0 degrees point

jfmichel:
When using `PieChartData`'s `startDegreeOffset`, it can happen a section spans across the 0 degrees point. 

In such a case, the aforementioned section will not be detected if touched on its "positive" side as the code assumes `fromDegree` will always be lower than `toDegree` (which would not be true here, angles being constrained in the [0:359] range).

This PR fixes this issue by considering the `relativeTouchAngle` which is the touch angle relative to the first section (whatever the `startDegreeOffset` might be). This approach keeps the code simple and now guarantees `fromDegree` will always be smaller than `toDegree`.

### DIFF
--- a/example/lib/pie_chart/samples/pie_chart_sample2.dart
+++ b/example/lib/pie_chart/samples/pie_chart_sample2.dart
@@ -37,6 +37,7 @@ class PieChart2State extends State {
                           }
                         });
                       }),
+                      startDegreeOffset: -90,
                       borderData: FlBorderData(
                         show: false,
                       ),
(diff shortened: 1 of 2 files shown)


example/lib/pie_chart/samples/pie_chart_sample2.dart
```diff
@@ -37,6 +37,7 @@ class PieChart2State extends State {
                           }
                         });
                       }),
+                      startDegreeOffset: -90,
```
imaNNeoFighT, line 40: We don't need to update this variable, because we need to update the sample screenshot too, 
please revert this change.
Thank you!

imaNNeoFighT:
Thanks for your contribution.

PR #364, opened 2020-06-01
Fix spot split when more than one in a row is null

tomwyr:
This PR fixes an issue where line chart fails to render if there's more than one `FlSpot` with null *x* or *y* in a row.

If there are two null spots in a row, then two empty lists will be added to `barList`. Next non-null spot will be added to the second one leaving an empty list in `barList` and causing rendering error.


imaNNeoFighT:
Great, 
thanks!

PR #366, opened 2020-06-02
Fix filling chart bounds

tomwyr:
Current logic assumes that first spot in the chart data is never null. In case it's missing e.g. *x* value then `maxX` is initialized with `null` and this part of code tries to compare spot value to it and throws NPE:
``` Dart
// Same for minX, minY and maxY
if (canModifyMaxX && spot.x > maxX) {
    maxX = spot.x;
}
```


imaNNeoFighT:
Good catch!
Thank you.

PR #417, opened 2020-08-19
Added the option to change the position of the text inside tooltip bo…

Uraskin:
…x, when overlay happened.

### DIFF
--- a/lib/src/chart/line_chart/line_chart_data.dart
+++ b/lib/src/chart/line_chart/line_chart_data.dart
@@ -1478,6 +1478,9 @@ class LineTouchTooltipData with EquatableMixin {
   /// Forces the tooltip to shift vertically inside the chart, if overflow happens.
   final bool fitInsideVertically;
 
+  /// Align text inside tooltipbox to center or to the corners, default 'true'
+  final bool tooltipTextAlignCenter;
+
   /// if [LineTouchData.handleBuiltInTouches] is true,
   /// [LineChart] shows a tooltip popup on top of spots automatically when touch happens,
   /// otherwise you can show it manually using [LineChartData.showingTooltipIndicators].
@@ -1500,6 +1503,7 @@ class LineTouchTooltipData with EquatableMixin {
     GetLineTooltipItems getTooltipItems,
     bool fitInsideHorizontally,
     bool fitInsideVertically,
+    bool tooltipTextAlignCenter,
   })  : tooltipBgColor = tooltipBgColor ?? Colors.white,
         tooltipRoundedRadius = tooltipRoundedRadius ?? 4,
         tooltipPadding = tooltipPadding ?? const EdgeInsets.symmetric(horizontal: 16, vertical: 8),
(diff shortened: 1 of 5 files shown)


imaNNeoFighT:
I don't understand the use-case of this feature,
Please write a descriptive description, and define the problem and then introduce your solution.
Thanks!

Uraskin:
> I don't understand the use-case of this feature,
> Please write a descriptive description, and define the problem and then introduce your solution.
> Thanks!

My use-case I ran into:  I setup option fitInsideHorizontally = true, then tooltipBox will be align to the right or left side. But text inside box always align to center, sometimes I need text be align to the left/right too.

imaNNeoFighT:
I understand the problem,
but this approach (having a `tooltipTextAlignCenter`) is not customizable enough.
I have a better approach to handle this in my head.

Inside the [LineTooltipItem](https://github.com/imaNNeoFighT/fl_chart/blob/master/repo_files/documentations/line_chart.md#linetooltipitem), we can add a property called textAlign (from [TextAlign](https://api.flutter.dev/flutter/dart-ui/TextAlign-class.html) enum), then we can use it in our painter, instead of hard-coded `Align.center` (check [line_chart_painter:1134](https://github.com/imaNNeoFighT/fl_chart/blob/master/lib/src/chart/line_chart/line_chart_painter.dart#L1134)).
It is so better for customizability.

Your use-case: 
then you can set `Alignment.left` for your first spot, and `Alignment.right` for your last spot.

Benefits: 
Now other people can use it freely and implement their use-cases without needing to add a new property or change it.

Let me know what you think.

Thanks for contributing!

Uraskin:
> I understand the problem,
> but this approach (having a `tooltipTextAlignCenter`) is not customizable enough.
> I have a better approach to handle this in my head.
> 
> Inside the [LineTooltipItem](https://github.com/imaNNeoFighT/fl_chart/blob/master/repo_files/documentations/line_chart.md#linetooltipitem), we can add a property called textAlign (from [TextAlign](https://api.flutter.dev/flutter/dart-ui/TextAlign-class.html) enum), then we can use it in our painter, instead of hard-coded `Align.center` (check [line_chart_painter:1134](https://github.com/imaNNeoFighT/fl_chart/blob/master/lib/src/chart/line_chart/line_chart_painter.dart#L1134)).
> It is so better for customizability.
> 
> Your use-case:
> then you can set `Alignment.left` for your first spot, and `Alignment.right` for your last spot.
> 
> Benefits:
> Now other people can use it freely and implement their use-cases without needing to add a new property or change it.
> 
> Let me know what you think.
> 
> Thanks for contributing!

I tried to use you're advise, but this did not work. Or maybe I have made something wrong.
So, property TextAling at line_chart_painter:1134, did not affect to positioned of tooltip text. If i changed it to left/right - nothing happened.
In my realisation I calculate position of tooltip text.

imaNNeoFighT:
Hmm, okay, 
what if we only use Alignment as a data holder?
I mean you can do your right calculations if the alignment is Alignment.right, and left calculations. if the alignment is left.
Or maybe we can have an enum for aligning it only left or right.

imaNNeoFighT:
I'm gonna close this. Because I didn't get any answer from you.
Thanks!

PR #443, opened 2020-09-08
[Pie Chart] New feature: added a badge widget in addition to title

gmaggio:
I needed a feature to add some kind of creative freedom in creating the title. So I added a way to add a widget to represent the title of the section. However, you can still have the default title as an option in addition to this widget. Having it as a widget also makes it intuitive to create and easy for novice programmers.

pie_chart_sample_3

## Main Changes:

- **PieChart:**
  - added child widget to CustomPaint to render the badge widgets
  - made default duration static to reuse externally for consistency
- **PieChartSectionData:**
  - title: changed default to `value`
  - added params for badge widget
- **PieChartPainter:**
  - added function to provide position offsets to parent
  - renamed `_drawTexts` to `_calculateOverlays` to better represent the various functions
- **Misc:**
  - Added `typedef` for position offsets map: `GetPositionOffsetsFunction`
  - Added test for PieChart's badge widget
  - Added sample source code for pie chart with badge widgets
  - Updated documentation


agungnursatria:
Cool!

imaNNeoFighT:
Hi dude, I am sorry for the delay.
I have just changed this PR's target branch to `dev`.
Today we have made some changes on dev, Now there is a conflict.
Can you please fix it?
Thanks, I will review it ASAP.

gmaggio:
Hey, thanks for the reponse 😄 
I have fixed the conflict. Looking forward to your review. 🙏 😇

ulusoyca:
@gmaggio Thanks for the PR. It is a huge contribution to my project also. I am waiting for the merge too. Looks like the conflicts are resolved. @imaNNeoFighT no pressure but you would make me happy too if you have time to test and possibly merge this :)

imaNNeoFighT:
I'm on it :)

imaNNeoFighT:
Hey @gmaggio.
First of all, I want to let you know that it is the most beautiful and complete pull request I have ever had.
I appreciate your great solution and great work, I bet you it helps people a lot. And they are going to be happy.
You did a great job!

There is only one minor suggestion from me. Now we don't need the `title` property in `PieChartSectionData` anymore and we can remove it. Because we can have a Text widget instead.

We can define a Text widget as a default value for the badge widget and show the index of each section inside it.
How do you think?

imaNNeoFighT:
Oh, I see you have used both of them at the same time.
Now you can skip my suggestion, but I would like to hear your idea.
Thanks!

imaNNeoFighT:
I think we can keep `title` to prevent make a breaking-change.
If you think we don't need it anymore, we can mark it deprecated!
I need your idea.
Thanks!

gmaggio:
Hi @imaNNeoFighT. Thank you for the complement. I really appreciate you looking into my contribution. 😇
In regards to the `title`, I think we should just **keep it** for a couple of reasons:

- It uses `CustomPainter`, which paints directly on the canvas during the painting phase.
  Though, to be honest, I'm not well versed on **canvas painting**, so my current assumption is that it is better than **widget manipulation** it terms of **performance**. Correct me if I'm wrong. 🤔
- The **_badge widget_** has yet to be **tested in various use cases and scenarios**.
  I had only tried it for my personal needs and it is working okay so far. But we need to allow time for others to use it as well for better feedback. Let's hope it is working as intended, though. 😁
- Keeping the `title` gives users **more options**.
  The `title` is technically more intuitive since it's more straight forward to use. Most people should be sufficient with this. The **_badge widget_**, on the other hand, allows customisation to the regular `title` but with an added effort, especially if animation is to be used. So we can leave the decision to the users to use whichever is suitable for their needs. And, of course, you can even use both of them at the same time, as you already know. 😉

These are only my point of views, though. Let me know what you think.
Cheers!

imaNNeoFighT:
All of your points make sense, and we will have both.
Thank you again!

PR #446, opened 2020-09-09
Fixing issue with only one section that is non-zero

makss68:
Fixing an issue when only one section of the pie chart has a non-zero value. In that case, the foundSectionData was returned null, because having only one section yields to `sectionAngle = 0`


imaNNeoFighT:
Hi.
I didn't understand the problem.
Can you please provide a reproducible code?

imaNNeoFighT:
It will be fixed in #615

PR #515, opened 2020-12-14
Improvement/context wrapper

imaNNeoFighT:
We implemented ContextWrapper class to proxy all canvas draw functions through it.
It makes our draw system testable.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,6 +21,14 @@ an issue before starting writing your code. This helps us and the
 community to discuss the issue and choose what is deemed to be the
 best solution.
 
+## Drawing architecture
+We have a *_chart_painter.dart class per each chart type. It draws elements into the Canvas.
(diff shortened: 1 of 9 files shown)


mhmdanas:
Why not use `Canvas` like before and mock it in tests using [mockito](https://pub.dev/packages/mockito)?

imaNNeoFighT:
Hmm I think you are right, can you please try it out and Implement a draw function's test?

mhmdanas:
I'll give it a stab tomorrow, but please bear in mind that I am not familiar with fl_chart's internals, so this might take more time than expected.

mhmdanas:
I just found an issue with my approach. We will have to mock `TextPainter` somehow, but that's not passed to the class; it's constructed inside of its methods. Maybe we should stick to your `CanvasWrapper` approach but combine it with `mockito` to reduce boilerplate. How about that?

PR #531, opened 2020-12-29
Improvement/slidetitle textdirection

NaarGes:
I'm using Persian language with this package and it didn't support RTL languages in side titles and tooltips. So I added optional `textDirection` property with `TextDirection.ltr` default value for side titles and tooltip.

### DIFF
--- a/lib/src/chart/bar_chart/bar_chart_data.dart
+++ b/lib/src/chart/bar_chart/bar_chart_data.dart
@@ -757,7 +757,7 @@ BarTooltipItem defaultBarTooltipItem(
     fontWeight: FontWeight.bold,
     fontSize: 14,
   );
-  return BarTooltipItem(rod.y.toString(), textStyle);
+  return BarTooltipItem(rod.y.toString(), textStyle, textDirection: TextDirection.ltr);
 }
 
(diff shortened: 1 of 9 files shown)


imaNNeoFighT:
We have added `textAlign` property in #601.
It will be great if you write a description.
I think that's why I didn't check this PR.

NaarGes:
I'm sorry that I forgot to write a description, I wrote it then.

I saw the commits of #601 pull request. It doesn't meet my needs. `TextAlign` is not the same as `TextDirection`. Text align indicates where the text is located within an element. But text direction specified which way the text flows, right-to-left or left-to-right. 

In my case I wanted to tag bottom titles with Persian dates like `۲۷ اسفند`, and it didn't displayed correctly without `textDirection` property.

imaNNeoFighT:
Got it. You are right!

imaNNeoFighT:
Can you please rebase your changes with the dev and let me know?
I highly suggest you make a new branch, because we have made a lot of changes since it opened.

imaNNeoFighT:
`textDirection` property added in SideTitles, AxisTitles and Tooltips in [0.35.0](https://github.com/imaNNeoFighT/fl_chart/releases/tag/0.35.0)

Issue #514, opened 2020-12-13
PieChart recent update 0.12.1 throws exception when pie slice is tapped


absar:
**Describe the bug**
PieChart recent update 0.12.1 throws exception when pie is tapped, however it works fine with version 0.12.0

**To Reproduce**
Tap Pie Chart, when the pie slice tries to enlarge exception is thrown, after that you cannot tap the slices anymore

**Screenshots**
======== Exception caught by widgets library 
The following assertion was thrown building PieChart(duration: 150ms, dirty, dependencies: [MediaQuery, _EffectiveTickerMode], state: _PieChartState#9a568(ticker inactive)):
RenderBox.size accessed beyond the scope of resize, layout, or permitted parent access. RenderBox can always access its own size, otherwise, the only object that is allowed to read RenderBox.size is its parent, if they have said they will. It you hit this assert trying to access a child's size, pass "parentUsesSize: true" to that child's layout().
'package:flutter/src/rendering/box.dart':
Failed assertion: line 1792 pos 13: 'debugDoingThisResize || debugDoingThisLayout ||
              (RenderObject.debugActiveLayout == parent && _size._canBeUsedByParent)'

Either the assertion indicates an error in the framework itself, or we should provide substantially more information in this error message to help you determine and fix the underlying cause.
In either case, please report this assertion by filing a bug on GitHub:
  https://github.com/flutter/flutter/issues/new?template=BUG.md

The relevant error-causing widget was: 
  PieChart
When the exception was thrown, this was the stack: 
#2      RenderBox.size.<anonymous closure> (package:flutter/src/rendering/box.dart:1792:13)
#3      RenderBox.size (package:flutter/src/rendering/box.dart:1805:6)
#4      _PieChartState._getChartSize (package:fl_chart/src/chart/pie_chart/pie_chart.dart:210:33)
#5      _PieChartState.badgeWidgets (package:fl_chart/src/chart/pie_chart/pie_chart.dart:160:23)
#6      _PieChartState.build (package:fl_chart/src/chart/pie_chart/pie_chart.dart:154:16)

**Versions**
 - Flutter version 1.22.5
 - FlChart version 0.12.1

imaNNeoFighT:
I couldn't reproduce this error.
Please send a reproducible code.
Thanks!

absar:
Hi here is the full code to reproduce. It's throwing error with both FlChart version 0.12.1 and 0.12.2, but it works fine with v0.12.0:

```
[...]

PR #535, opened 2021-01-06
Bugfix/piechart resize

xinoxapps:
Fixes issue #514 which can occur when the PieChart resizes in a Sliver.


bdlukaa:
When can this be merged?

imaNNeoFighT:
Now!

bdlukaa:
Now a different error is thrown. `Another exception was thrown: A build function returned null.`.

Instead of `return null;`, it should be `return SizedBox();`

bdlukaa (commented):
`A build function returned null.` `badgeWidgets` is returning null

PR #578, opened 2021-02-18
Add extraVerticalOffset value to VerticalLine

jseminck:
Hi. First of all thanks for the amazing library, it's really awesome!

I have a use-case where I need `VerticalLine` elements to draw outside of the container that the chart is rendered in, all the way to the top of the screen. The best way I thought to achieve this was to add extra property to the `VerticalLine` and use this when drawing the vertical line.

So now we can draw from y location `0 - 300` which will draw a vertical line from 300 pixels above the chart container until the bottom of the container.

What do you think about this idea? Also, please suggest alternative naming for the variable, if you think there is a better name.

### DIFF
--- a/lib/src/chart/line_chart/line_chart_data.dart
+++ b/lib/src/chart/line_chart/line_chart_data.dart
@@ -1064,6 +1064,10 @@ class VerticalLine extends FlLine with EquatableMixin {
   /// Draws a text label over the line.
   final VerticalLineLabel label;
 
+  /// Additional vertical offset, what can be used to draw outside of the chart area, for example if
+  /// the vertical line should go all the way to the top of the screen.
+  final double extraVerticalOffset;
+
   /// [LineChart] draws vertical lines from bottom to top side of the chart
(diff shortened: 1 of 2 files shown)


jseminck:
Hi @imaNNeoFighT , do you have some time to look at this PR and give your thoughts? If you approve of the approach I will do my best to add tests, update changelog and documentations, and anything else what is necessary for it to be approved and merged.

I'll try to explain the reasoning a bit more. We have a chart and I have drawn a rectangle around the container in which it is rendered. So the vertical line ends at the boundary of the container. But we want it to extend all the way to the top instead. For different reasons it is not easy to refactor the layout of the widgets into a Stack (for example) where the container has full height. 

image

imaNNeoFighT:
Hi @jseminck.
First of all, excuse me for the late reply.
I was busy as a bee.
I think this is a particular use-case. Also, it is a little weird that widgets draw something out of their boundaries.
I suggest you fork the repository and maintain this feature and use it in your project.

PR #604, opened 2021-03-15
Setting start/end point of the TouchLine

terryl1900:
Hi @imaNNeoFighT , I'm adding a feature to allow configuring the start and end point of the TouchLine of LineChart. See the image for the use case. I also added an example.

This method will replace the `fullHeightTouchLine` parameter with `getTouchLineEnd`, so it's a breaking change.

Screenshot 2021-03-15 at 16 01 23
Screenshot 2021-03-15 at 16 01 33


lib/src/chart/line_chart/line_chart_data.dart
```diff
@@ -1330,14 +1336,18 @@ class LineTouchData extends FlTouchData with EquatableMixin {
     LineTouchTooltipData? touchTooltipData,
     GetTouchedSpotIndicator? getTouchedSpotIndicator,
     double? touchSpotThreshold,
-    bool? fullHeightTouchLine,
     bool? handleBuiltInTouches,
+    GetTouchLineY? getTouchLineStart,
+    GetTouchLineY? getTouchLineEnd,
+    bool? touchLineEndAtDot,
```
imaNNeoFighT, line 1342: I suggest `preventOverlapWithDot` instead of `touchLineEndAtDot`.
(The functionality must be negated)

lib/src/chart/line_chart/line_chart_data.dart
```diff
@@ -1410,6 +1429,16 @@ List<TouchedSpotIndicatorData> defaultTouchedIndicators(
   }).toList();
 }
 
+/// By default line starts from the bottom of the chart.
+double defaultGetTouchLineStart(LineChartBarData barData, int spotIndex) {
+  return -double.infinity;
```
imaNNeoFighT, line 1434: What is the purpose of minus here?

lib/src/chart/line_chart/line_chart_painter.dart
```diff
@@ -297,19 +298,26 @@ class LineChartPainter extends AxisChartPainter<LineChartData>
...
-      final lineEnd =
-          data.lineTouchData.fullHeightTouchLine ? top : touchedSpot + Offset(0, dotHeight / 2);
+      final lineStartY =
+          min(data.maxY, max(data.minY, data.lineTouchData.getTouchLineStart(barData, index)));
+      final lineEndY =
+          min(data.maxY, max(data.minY, data.lineTouchData.getTouchLineEnd(barData, index)));
+      final lineStart = Offset(touchedSpot.dx, getPixelY(lineStartY, chartViewSize));
+      var lineEnd = Offset(touchedSpot.dx, getPixelY(lineEndY, chartViewSize));
+      if (data.lineTouchData.touchLineEndAtDot) {
+        if (lineStart.dy < lineEnd.dy) {
```
imaNNeoFighT, line 308: What if `lineStart` is greater or equals to `lineEnd`?
I think we need to throw an exception if `getTouchLineEnd` provided a value lower than `getTouchLineEnd`.

lib/src/chart/line_chart/line_chart_painter.dart
```diff
@@ -297,19 +298,26 @@ class LineChartPainter extends AxisChartPainter<LineChartData>
...
-      /// Draw to top or to the touchedSpot
-      final lineEnd =
-          data.lineTouchData.fullHeightTouchLine ? top : touchedSpot + Offset(0, dotHeight / 2);
+      final lineStartY =
+          min(data.maxY, max(data.minY, data.lineTouchData.getTouchLineStart(barData, index)));
+      final lineEndY =
+          min(data.maxY, max(data.minY, data.lineTouchData.getTouchLineEnd(barData, index)));
+      final lineStart = Offset(touchedSpot.dx, getPixelY(lineStartY, chartViewSize));
+      var lineEnd = Offset(touchedSpot.dx, getPixelY(lineEndY, chartViewSize));
+      if (data.lineTouchData.touchLineEndAtDot) {
```
imaNNeoFighT, line 307: What if `lineStart` is greater or equals to `lineEnd`?

imaNNeoFighT: Got it!

lib/src/chart/line_chart/line_chart_data.dart
imaNNeoFighT: Got it!

lib/src/chart/line_chart/line_chart_data.dart
imaNNeoFighT: BTW It has a lot of edge-cases.
For example, determine user getTouchLineStart start to -double.infinity, and getTouchLineEnd to +double.infinity while this flag is true.
And we decrease the end of the line by circle size which it passed over the circle.

imaNNeoFighT: Please remove this flag, and if you want we can talk about it in a separate PR.
Because there are a lot of edge-cases that I can imagine.
Thanks!

imaNNeoFighT:
Hello?

lib/src/chart/line_chart/line_chart_data.dart
terryl1900: Sorry for the delay. I was out a few days.

Yeah, I think what you said makes sense. Removed this parameter. Though I would like to keep the current behavior that the line end doesn't overlap with the dot. Now, this is done by checking whether the line ends within the dot.

example/lib/line_chart/line_chart_page3.dart
```diff
@@ -20,6 +21,17 @@ class LineChartPage3 extends StatelessWidget {
...
+              height: 52,
+            ),
+            const Text(
+              'LineChart (positive and negative values)',
+              style: TextStyle(fontSize: 32, fontWeight: FontWeight.bold, color: Colors.black),
+            ),
+            const SizedBox(
+              height: 52,
+            ),
+            LineChartSample9(),
```
imaNNeoFighT, line 34: Wrap them with a ListView (instead of current Column)

imaNNeoFighT: Or I can do that myself after we merge.
Because I want to show the sample in our samples section.

lib/src/chart/line_chart/line_chart_data.dart
imaNNeoFighT: No problem.
Great idea. Thanks, I'm going to merge it.

imaNNeoFighT:
Thanks for contributing!
Hope to see more from you.

PR #632, opened 2021-04-07
Migrated the third PieChart example to null safety

4amVim:
As null safety is now de-facto branch most users are using, the examples should work on it out-of-the-box.


4amVim:
Hey, 

Apologies for pinging you @imaNNeoFighT, but any reason for not merging this PR?

Issue #260, opened 2020-03-24
rotate text in BarTooltipItem


alexodus:
**Is your feature request relasted to a problem? Please describe.**
If number is large it, overlaps over the bar

**Describe the solution you'd like**
I'd like to have an rotateAngle property also for BarTooltipItem

**Additional context**
![image](https://user-images.githubusercontent.com/2721510/77415743-31379d00-6dc3-11ea-88f4-08a4da5349a9.png)

imaNNeoFighT:
Thanks for reporting,
we will add it ASAP.
Thanks!

alexodus:
Hi @imaNNeoFighT , are there any news/progress?

imaNNeoFighT:
Hi, unfortunately, we have more priority things todo (priority is determined by the count of thumbs up).
Also, notice that you can contribute and create a Pull Request and help us.
Thanks!

imaNNeoFighT:
Hi. We have just added `rotateAngle` property in  [LineTouchTooltipData](https://github.com/imaNNeoFighT/fl_chart/blob/master/repo_files/documentations/line_chart.md#linetouchtooltipdata), [BarTouchTooltipData](https://github.com/imaNNeoFighT/fl_chart/blob/master/repo_files/documentations/bar_chart.md#bartouchtooltipdata) and [ScatterTouchTooltipData](https://github.com/imaNNeoFighT/fl_chart/blob/master/repo_files/documentations/scatter_chart.md#scattertouchtooltipdata).

Check it out in [0.36.2](https://github.com/imaNNeoFighT/fl_chart/releases/edit/0.36.2)

PR #635, opened 2021-04-08
Allow parallel tooltip on bar chart

marc921:
I believe this PR should solve this issue : <https://github.com/imaNNeoFighT/fl_chart/issues/260>

Allows tooltip to be written parallel to bar like this:
Capture d’écran 2021-04-08 à 17 34 53


imaNNeoFighT:
I think we must implement a way to allow users to set an angle for showing the tooltip.
That solves all problems with angle.

BTW Your solution is not working correctly.

Check image below:

imaNNeoFighT:
We are going to solve #260 problem in #679.

PR #642, opened 2021-04-19
Add new fields to RadarChartData: set title's angle, callback for tap…

zeienko-vitalii:
It occurs that it's very useful to have an ability to:
- rotate title outside of the `RadarChart`
- get tap event by pressing on title
- changing tick's title

The changes are introduced in following classes: `RadarChartData` and `RadarChartPainter`.
New fields in `RadarChartData` that are used in `RadarChartPainter`:
- `double? titleAngle`
- `GetTickTitleFunction? getTickTitle`, where `GetTickTitleFunction` is a typedef of this function: `String Function(int index, double tick)`
- `Function(String)? onTitleTap`


lib/src/chart/radar_chart/radar_chart_painter.dart
```diff
@@ -183,9 +187,9 @@ class RadarChartPainter extends BaseChartPainter<RadarChartData> {
         centerX + threshold * radius * xAngle,
         centerY + threshold * radius * yAngle,
       );
+      _titlesPosition[index] = featureOffset;
       canvasWrapper.translate(featureOffset.dx, featureOffset.dy);
-      canvasWrapper.rotate(angle * index);
-
+      canvasWrapper.rotate(data.titleAngle ?? (angle * index));
```
imaNNeoFighT, line 192: I think we can make a typedef (like what you did for getTickTitle) and make data.titleAngle customizable per each index (instead of one titleAngle value)

lib/src/chart/radar_chart/radar_chart_painter.dart
```diff
@@ -107,7 +108,7 @@ class RadarChartPainter extends BaseChartPainter<RadarChartData> {
         canvasWrapper.drawCircle(centerOffset, tickRadius, _tickPaint);
         _ticksTextPaint
           ..text = TextSpan(
-            text: tick.toStringAsFixed(1),
+            text: data.getTickTitle?.call(index, tick) ?? tick.toStringAsFixed(1),
```
imaNNeoFighT, line 111: As a convention, we move our default logic into a default implementation of that typedef, and we set the default implementation in the constructor (If the user didn't customize it).
For example please check [here](https://github.com/imaNNeoFighT/fl_chart/blob/2d42fdb0969eb3e3a8eac39637ea96083fe28b94/lib/src/chart/line_chart/line_chart_data.dart#L731).

lib/src/chart/radar_chart/radar_chart_painter.dart
zeienko-vitalii: Will do

lib/src/chart/radar_chart/radar_chart_painter.dart
zeienko-vitalii: Got it, I'll change it. Thank you!

lib/src/chart/radar_chart/radar_chart_painter.dart
```diff
@@ -246,6 +250,7 @@ class RadarChartPainter extends BaseChartPainter<RadarChartData> {
 
   RadarTouchedSpot? handleTouch(
       PointerEvent touchInput, Size size, PaintHolder<RadarChartData> holder) {
+    _triggerTitleTap(touchInput.down, touchInput.localPosition, holder);
```
imaNNeoFighT, line 253: `handleTouch` function gets called in every touch event (such as move touch).
So this is not the best way to handle this.
We need to change our approach to handle this kind of custom touch. (I know you need it in your use-case) But we cannot handle this as an instant solution (because we need to consider our long-term approach).
So please revert these codes for now, and open an issue to follow it up.

Thanks!

zeienko-vitalii: @imaNNeoFighT please take a look at the latest changes

zeienko-vitalii: @imaNNeoFighT it'd be great if you check the latest changes

imaNNeoFighT: I don't find where you are checking _TransformedPointerUpEvent?

There is only one if block which is :
```
if (isDownEvent || _titlesPosition.isEmpty) return;
```

zeienko-vitalii: > I don't find where you are checking _TransformedPointerUpEvent?
> 
> There is only one if block which is :
> 
> ```
> if (isDownEvent || _titlesPosition.isEmpty) return;
> ```

Yes, actually it is, @imaNNeoFighT

zeienko-vitalii: @imaNNeoFighT , can I leave it for merge or should I remove it?

imaNNeoFighT: Please remove `_triggerTitleTap` to speed up the merge process.
We can talk about it separately.

Sorry for my delay. I was busy.

imaNNeoFighT:
Closed due to inactivity for a long time.

PR #644, opened 2021-04-22
Fix getNearestTouchedSpot

### LINKED ISSUE #641: Show number tooltip on the closest value to the cursor even when treshold is set
_I'm sorry if this is a duplicate, I've looked trough the issues quickly, but I might have missed a similar one._

My problem is that when the user taps the graph, the number shows up to the right of the cursor, at the distance of the treshold.\
Treshold works great when there are only a few values in a graph, meaning it snaps to the closest value in that range, working both in the right and left directions.
**However, when there are many values, the tooltip appears way off to the side.**\
I understand this may be beneficial in cases where you have to avoid showing the value below the user's finger, but in our case, this is more annoying than is useful, we're getting bug reports of it.\
Also, like this, it's hard to check the values right at the start of the graph, or even impossible on smaller screens.
#### Demonstration:
image

#### Proposal:
There could be an option to turn on "closest-snap", meaning the tooltip would show for the closest value to the cursor in the range of the treshold on the graph.

PerLycke:
Previously it returned the first occurrence of a spot within the threshold, and not the nearest, which I understand is the purpose of this function.

With the updated function the user is able to set the threshold to a big number (like double.maxFinite) and get the nearest spot from this function. This behaviour (clicking anywhere and get the nearest spot) is seen in e.g. charts_flutter.

AFAIK this shouldn't break anything for existing users, just fix / add the above functionality.

Fixes #641, #645.

### DIFF
--- a/lib/src/chart/line_chart/line_chart_painter.dart
+++ b/lib/src/chart/line_chart/line_chart_painter.dart
@@ -1416,8 +1416,12 @@ class LineChartPainter extends AxisChartPainter<LineChartData> {
   }
 
   /// find the nearest spot base on the touched offset
-  LineBarSpot? _getNearestTouchedSpot(Size viewSize, Offset touchedPoint, LineChartBarData barData,
-      int barDataPosition, PaintHolder<LineChartData> holder) {
+  LineBarSpot? _getNearestTouchedSpot(
+      Size viewSize,
+      Offset touchedPoint,
+      LineChartBarData barData,
+      int barDataPosition,
+      PaintHolder<LineChartData> holder) {
     final data = holder.data;
     if (!barData.show) {
       return null;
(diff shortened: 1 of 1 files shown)


lib/src/chart/line_chart/line_chart_painter.dart
```diff
@@ -1416,26 +1416,44 @@ class LineChartPainter extends AxisChartPainter<LineChartData> {
...
-    for (var i = 0; i < barData.spots.length; i++) {
-      final spot = barData.spots[i];
-      if (spot.isNotNull()) {
-        if ((touchedPoint.dx - getPixelX(spot.x, chartViewSize, holder)).abs() <=
-            data.lineTouchData.touchSpotThreshold) {
-          return LineBarSpot(barData, barDataPosition, spot);
+    final sortedSpots = <FlSpot>[];
+    double? smallestDistance;
+
+    for (var e in barData.spots) {
```
imaNNeoFighT, line 1435: Everything is well implemented and working great.
I have only one minor comment. `e` is not descriptive enough. please rename it to something like: `spot`.

Thanks!

imaNNeoFighT:
I reviewed it and everything is okay to be merged.
Please resolve my conversation. Then fix the check style problem (Check out our [contribution guideline](https://github.com/imaNNeoFighT/fl_chart/blob/master/CONTRIBUTING.md) to fix it.

Thanks!

imaNNeoFighT:
Hellooo?

PerLycke:
I never got to use the plugin but sure, i will fix this asap (prob tomorrow). Or you do (as it's a trivial change) and merge the changes, I don't mind.

imaNNeoFighT:
> I never got to use the plugin but sure, i will fix this asap (prob tomorrow). Or you do (as it's a trivial change) and merge the changes, I don't mind.

Thanks for responding.
I will be happy if we fix this under your name. Because you (community) are the main value of the open source projects.

BTW if you are busy or you feel uncomfortable, we can use your code.

Thanks

PerLycke:
@imaNNeoFighT I fixed it just now 👍

imaNNeoFighT:
Thanks man.
There is a formatting issue.
Please run `make format` and If you can combine all changes into one commit

PerLycke:
I formatted the code in question for this pull request and combined it all into one commit. Should be all good now?

PerLycke:
I don't get this check. The code I've changed / added is properly formatted (using Dartfmt), but as far as I can see the check fails because the file in whole is not formatted? I'm supposed to format the rest of the (by me) untouched code, which is out of scope from my fix, too? That would also make it harder to find my specific fix among all formatting changes.

@imaNNeoFighT IMO my code is properly formatted and this check is failing due to everything else not being properly formatted and that should be fixed in a separate commit not related to this PR. What do you think?

imaNNeoFighT:
No, still there is a problem.
please make sure you can run `make checkstyle` without any error. If there was any problem you can run `make format` to fix those issues.
BTW before you push make sure that `make checkstyle` runs without any error.

imaNNeoFighT:
We still have this bug.
Can you please rebase your code again and let me know?

imaNNeoFighT:
I re-implemented it in #850. Like exactly what you did.

PR #666, opened 2021-05-06
specify that it uses the short scale for bilions

elpablete:
There are two definitions for the number billion (in short scale and in long scale) as stated in https://en.wikipedia.org/wiki/Billion


imaNNeoFighT:
Thanks for mentioning!

Issue #714, opened 2021-07-08
Add borders to BarChart bars


eliseyOzerov:
**Is your feature request relasted to a problem? Please describe.**
In our BarChart implementation, a part of the current daily bar (weekly chart where each bar is a day) doesn't represent a value, hence it's white. Because the background of the app is also white, there's no way to distinguish the empty bar from the background. That's mandated by our design doc so I can't just change the color.

**Describe the solution you'd like**
We'd like to have an option to add borderData for each BarChartRodData in the BarChart.

**Describe alternatives you've considered**
As I said above, the only other option I see to solve our problem would be to change the color of that particular bar segment, but that's currently not an option.

imaNNeoFighT:
Does `backDrawRodData` help you?

Check the below sample.

### Some Samples
----
##### Sample 1 ([Source Code](https://github.com/imaNNeoFighT/fl_chart/blob/master/example/lib/bar_chart/samples/bar_chart_sample1.dart))
<img src="https://github.com/imaNNeoFighT/fl_chart/raw/master/repo_files/images/bar_chart/bar_chart_sample_1.gif" width="300" >

imaNNeoFighT:
Hi.
We have just added `borderSide` property in [BarChartRodData](https://github.com/imaNNeoFighT/fl_chart/blob/master/repo_files/documentations/bar_chart.md#barchartroddata) and [BarChartRodStackItem](https://github.com/imaNNeoFighT/fl_chart/blob/master/repo_files/documentations/bar_chart.md#barchartrodstackitem) to draw strokes around each bar and rod stack items in [0.36.4](https://github.com/imaNNeoFighT/fl_chart/releases/tag/0.36.4).
check it out!

Thanks @samiazar, hope to see more from you :)

PR #725, opened 2021-07-23
Feature: add border for barchart

samiazar:
* **IMPROVEMENT** Added `borderSide` property in [BarChartRodData](https://github.com/imaNNeoFighT/fl_chart/blob/master/repo_files/documentations/bar_chart.md#BarChartRodData) and [BarChartRodStackItem](https://github.com/imaNNeoFighT/fl_chart/blob/master/repo_files/documentations/bar_chart.md#BarChartRodStackItem) to draw strokes around each bar and rod stack items, #714.


lib/src/utils/utils.dart
```diff
@@ -82,6 +82,25 @@ BorderRadius? normalizeBorderRadius(BorderRadius? borderRadius, double width) {
...
+  }
+
+  double borderWidth;
+  if (borderSide.width > width / 2) {
+    borderWidth = width / 2.toDouble();
+  } else {
+    borderWidth = borderSide.width;
+  }
+
+  return BorderSide(color: borderSide.color, width: borderWidth, style: borderSide.style);
```
imaNNeoFighT, line 101: You can use borderSide.copy() here

lib/src/chart/bar_chart/bar_chart_painter.dart
```diff
@@ -317,6 +326,9 @@ class BarChartPainter extends AxisChartPainter<BarChartData> {
               canvasWrapper.drawRRect(barRRect, _barPaint);
               canvasWrapper.restore();
             }
+            // draw border stroke for each stack item
+            _drawRodStacksBorderStroke(
```
imaNNeoFighT, line 330: Above we iterated over stack items to draw them, why we iterated over them again inside this function?
I think we can use one iteration over stack items at all.

imaNNeoFighT:
Everything looks good to me.
Thanks Sami!

lib/src/utils/utils.dart
samiazar: Thank you for the good tip.

lib/src/chart/bar_chart/bar_chart_painter.dart
samiazar: You're right, I've changed the code.

Issue #724, opened 2021-07-22
Give possibility to align vertical chart marks to the right or to the left


ruan65:
If there is such a feature, tell me please how to do so, please. I did not find it.

<img width="462" alt="Screenshot 2021-07-22 at 17 15 35" src="https://user-images.githubusercontent.com/4295256/126655031-6db42ee6-a43f-4a36-b9ec-7373e3956e4f.png">

I have tried to align by playing with string length, but I failed (

ruan65:
I forked the repo and found the way to do that.

The PR will be soon.

ruan65:
Pull request created

https://github.com/imaNNeoFighT/fl_chart/pull/730

PR #730, opened 2021-07-27
Left and Right side titles alignment to the chart viewport edge

ruan65:
Often designers create prototypes for screens containing a chart drawing the chart Y marks aligned with screen content to the screen left and right paddings. This code allows doing that.

It is enough to add one param to the SideTitles constructor

SideTitles _getRightForecastChartTitles(double yInterval) => SideTitles(
...
alignRightTitlesToRight: true,
);

Also, I added an example (which can be deleted if it does not suit)

Screenshot of the example

Simulator Screen Shot - iPhone 12 Pro Max - 2021-07-27 at 06 05 43

Issue:

https://github.com/imaNNeoFighT/fl_chart/issues/724

### DIFF
--- a/example/ios/Runner.xcodeproj/project.xcworkspace/contents.xcworkspacedata
+++ b/example/ios/Runner.xcodeproj/project.xcworkspace/contents.xcworkspacedata
@@ -2,6 +2,6 @@
 <Workspace
    version = "1.0">
    <FileRef
-      location = "group:Runner.xcodeproj">
+      location = "self:">
    </FileRef>
 </Workspace>
--- a/example/lib/line_chart/line_chart_page5.dart
+++ b/example/lib/line_chart/line_chart_page5.dart
@@ -1,0 +1,57 @@
+import 'package:example/line_chart/samples/line_chart_sample11.dart';
+import 'package:flutter/material.dart';
+
+class LineChartPage5 extends StatelessWidget {
+  @override
+  Widget build(BuildContext context) {
+    return Container(
+      color: Colors.white,
+      child: Center(
+        child: ListView(
+          padding: EdgeInsets.symmetric(horizontal: 28, vertical: 18),
+          children: <Widget>[
+            const Text(
+              'LineChart (Left titles alignment to the left and '
+              'Right titles alignment to the right example)',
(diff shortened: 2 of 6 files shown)


imaNNeoFighT:
Hi. First of all thanks for contributing. I really appreciate.
BTW we need to solve this problem more generic. I think it is better to have a [TextAlign](https://api.flutter.dev/flutter/dart-ui/TextAlign-class.html) property inside the SideTitles. It can be `TextAlign.center`, `TextAlign.left` and `TextAlign.right`.
Then you can handle the logic for these values in right and left sideTitles.

ruan65:
> Hi. First of all thanks for contributing. I really appreciate.
> BTW we need to solve this problem more generic. I think it is better to have a [TextAlign](https://api.flutter.dev/flutter/dart-ui/TextAlign-class.html) property inside the SideTitles. It can be `TextAlign.center`, `TextAlign.left` and `TextAlign.right`.
> Then you can handle the logic for these values in right and left sideTitles.

Ok, I will rework it soon.

imaNNeoFighT:
Thanks!

dumazy:
> > Hi. First of all thanks for contributing. I really appreciate.
> > BTW we need to solve this problem more generic. I think it is better to have a [TextAlign](https://api.flutter.dev/flutter/dart-ui/TextAlign-class.html) property inside the SideTitles. It can be `TextAlign.center`, `TextAlign.left` and `TextAlign.right`.
> > Then you can handle the logic for these values in right and left sideTitles.
> 
> Ok, I will rework it soon.

Any progress on this? Looking for this exact solution as @imaNNeoFighT suggests it. If not, I'll work on a new PR for this.

kzlakowski:
> > > Hi. First of all thanks for contributing. I really appreciate.
> > > BTW we need to solve this problem more generic. I think it is better to have a [TextAlign](https://api.flutter.dev/flutter/dart-ui/TextAlign-class.html) property inside the SideTitles. It can be `TextAlign.center`, `TextAlign.left` and `TextAlign.right`.
> > > Then you can handle the logic for these values in right and left sideTitles.
> > 
> > 
> > Ok, I will rework it soon.
> 
> Any progress on this? Looking for this exact solution as @imaNNeoFighT suggests it. If not, I'll work on a new PR for this.

I've already done it #849

PR #737, opened 2021-08-05
Bugfix: colorStops calculation

1step2hell:


imaNNeoFighT:
There is a problem.
If we have 1 point, we divide 1.0 / 0.0 which leads to "dividing by zero exception".

1step2hell:
We shouldn't use gradient colors if there is only one point. I searched everywhere function "getSafeColorStops" was invoked, you have checked colors.length==1(but haven't check colors.length==0).  Still, I agree with you, we should make this function more robust. So I add a check and throw an exception. How do you think?

example/lib/line_chart/samples/line_chart_sample5.dart
```diff
@@ -144,12 +144,17 @@ class LineChartSample5 extends StatelessWidget {
 
 /// Lerps between a [LinearGradient] colors, based on [t]
 Color lerpGradient(List<Color> colors, List<double> stops, double t) {
+  if (colors.isEmpty) {
+    return Colors.green;
```
imaNNeoFighT, line 148: I think it is better to throw an Exception instead of returning a hard-coded value.

lib/src/chart/line_chart/line_chart_data.dart
```diff
@@ -697,7 +697,7 @@ typedef GetDotColorCallback = Color Function(FlSpot, double, LineChartBarData);
 /// If there is one color in [LineChartBarData.colors], it returns that color,
 /// otherwise it returns the color along the gradient colors based on the [xPercentage].
 Color _defaultGetDotColor(FlSpot _, double xPercentage, LineChartBarData bar) {
-  if (bar.colors.isEmpty || bar.colors.isEmpty) {
+  if (bar.colors.isEmpty) {
     return Colors.green;
```
imaNNeoFighT, line 701: And I think here is better to throw an exception (I know it is out of this PR's scope)

lib/src/chart/line_chart/line_chart_data.dart
```diff
@@ -710,7 +710,7 @@ Color _defaultGetDotColor(FlSpot _, double xPercentage, LineChartBarData bar) {
 /// otherwise it returns the color along the gradient colors based on the [xPercentage] in a darker mode.
 Color _defaultGetDotStrokeColor(FlSpot spot, double xPercentage, LineChartBarData bar) {
   Color color;
-  if (bar.colors.isEmpty || bar.colors.isEmpty) {
+  if (bar.colors.isEmpty) {
     color = Colors.green;
```
imaNNeoFighT, line 714: And here.

imaNNeoFighT:
> We shouldn't use gradient colors if there is only one point. I searched everywhere function "getSafeColorStops" was invoked, you have checked colors.length==1(but haven't check colors.length==0). Still, I agree with you, we should make this function more robust. So I add a check and throw an exception. How do you think?

I agree with throwing an exception.
Well done. I'm going to merge this PR after those minor conversations.

imaNNeoFighT:
Also, don't forget to run `make checkstyle` and `make format` to fix format problems. Also please run `make runTests` to make sure all unit tests are passing.

example/lib/line_chart/samples/line_chart_sample5.dart
1step2hell: done

lib/src/chart/line_chart/line_chart_data.dart
1step2hell: done

lib/src/chart/line_chart/line_chart_data.dart
1step2hell: done

imaNNeoFighT:
Thanks for contributing!

PR #740, opened 2021-08-11
Refactoring line_chart_sample_1

bakatsuyuki:
This library is fantastic.  But I thought the samples were room for improvement because there were many classes and arguments in this library.

First, that developers who use this library should want to know about this library.
So they don't have to see codes unrelated to this library(For example, StatefulWidget, AspectRatio, and so on).
In other words, I think the codes about this library are better if on the top of the sample.

Then, I think the sample codes are better if divided into smaller functions because this library can do many excellent jobs.

I tried to refactor it. I think it's easier to read than before. What do you think?


imaNNeoFighT:
Thanks! I appreciate it.
Also please help me with refactoring other chart samples.

PR #743, opened 2021-08-17
implementing overlap title on left line chart feature

Letalus:
Hey I needed a chart extension where I was able to display the y axis title over the chart as shown in the screenshot.
I forked your package and did the minor adjustments within the package.

I wanted to ask, if you would be interested in including this feature within your package.

I know, currently this feature hasnt yet covered all cases like bottom, top and right title overlapping because I only need the left side.

In case you would be interested I could also implement those cases.

Kind regards,
Chris


imaNNeoFighT:
Hi. I think you forget to attach the screenshot.

Letalus:
IMG_6688BB2AEAF6-1

yeap indeed, sorry for the inconvience.

imaNNeoFighT:
I didn't understand the use-case. Do you need to put titles inside the chart? (instead of outside)

Issue #603, opened 2021-03-15
bottomTitles - Last title overflowed


dospringer:
**Describe the bug**
The problem is, that the last title on the x-axis on SideTitle overflowed. Is there a way that the last title don't overflowed?

**Screenshot**
<img width="1220" alt="Bildschirmfoto 2021-03-15 um 14 28 25" src="https://user-images.githubusercontent.com/71374289/111160875-d3373300-859a-11eb-94c3-b903dc55ee17.png">

**Versions**
 Flutter: (Channel master, 2.1.0-13.0.pre.64)
fl_chart:  ^0.20.1

**To Reproduce**
```dart
import 'package:flutter/foundation.dart';
import 'package:flutter/material.dart';
import 'package:fl_chart/fl_chart.dart';

void main() => runApp(MyApp());

class MyApp extends StatelessWidget {
  @override
  Widget build(BuildContext context) {
    return MaterialApp(
      showPerformanceOverlay: false,
      theme: ThemeData(
        primaryColor: const Color(0xff262545),
        primaryColorDark: const Color(0xff201f39),
        brightness: Brightness.dark,
      ),
      home: const MyHomePage(title: 'fl_chart'),
    );
  }
}

class MyHomePage extends StatefulWidget {
  const MyHomePage({Key key, this.title}) : super(key: key);
  final String title;

  @override
  _MyHomePageState createState() => _MyHomePageState();
}

class _MyHomePageState extends State<MyHomePage> {
  final _controller = PageController(initialPage: 0);
  final _pages = [
    LineChartPage(),
  ];

  @override
  Widget build(BuildContext context) {
    return Scaffold(
      body: SafeArea(
        child: PageView(
          physics: kIsWeb
              ? NeverScrollableScrollPhysics()
              : AlwaysScrollableScrollPhysics(),
          controller: _controller,
          children: _pages,
        ),
      ),
    );
  }
}

class LineChartPage extends StatelessWidget {
  @override
  Widget build(BuildContext context) {
    return Container(
      color: const Color(0xff262545),
      child: ListView(
        children: <Widget>[
          const Align(
            alignment: Alignment.centerLeft,
            child: Padding(
              padding: EdgeInsets.only(
                left: 36.0,
                top: 24,
              ),
            ),
          ),
          const SizedBox(
            height: 8,
          ),
          Padding(
            padding: const EdgeInsets.only(
              left: 28,
              right: 28,
            ),
            child: LineChartSample1(),
          ),
        ],
[...]

PR #781, opened 2021-09-29
Add option to hide overflow titles

petercinibulk:
This PR addresses https://github.com/imaNNeoFighT/fl_chart/issues/603

This is a simple feature to add the option to hide top and bottom titles IF they overflow the left or right borders of the chart.
All it does is check if the text will be drawn outside the bounds on the left and right sides of the canvas area. If the title does overflow the boundary, then it is not painted.

I could not figure out a way to write a test for this but would gladly update this if anyone has any ideas on a good test.

Usage:
```
topTitles: SideTitles(
          showTitles: true,
          hideTitleOnOverflow: true,  
          ...
),
bottomTitles: SideTitles(
          showTitles: true,
          hideTitleOnOverflow: true,  
          ...
)
```


imaNNeoFighT:
Hi
Thanks for contributing.
I think it doesn't solve the problem. we are just ignoring it (by hiding the first and last titles). 
For example in the issue that you've mentioned, as we need to show 3 bottom titles, we can't only hide the last title.

Bildschirmfoto 2021-03-15 um 14 28 25

Issue #811, opened 2021-11-09
Bar chart -  leftTitles are not showing up properly


svbala99:
**Description the bug**
In bar chart, left titles are not showing proper intervals when the value becomes fractional.

- When the fed plot values go beyond zero (negative), the leftTitles are not being properly shown, shows just min, mid, max intervals on Y axis at times, and just min value alone sometimes. 
- We tried getting the value from getTitles for left side titles as below:-
```
getTitles: (value) {
       print(value);
       return value.toString();
}),
```
- But for some set of plot values, getTitles returns the values 0 and 1 alone.

**To Reproduce**
```
// main.dart
import 'package:fl_chart/fl_chart.dart';
import 'package:flutter/material.dart';

void main() {
  runApp(const MyApp());
}

class MyApp extends StatelessWidget {
  const MyApp({Key? key}) : super(key: key);

  @override
  Widget build(BuildContext context) {
    return MaterialApp(
      title: 'Bar chart',
      theme: ThemeData(
        primarySwatch: Colors.blue,
      ),
      home: const MyHomePage(title: 'Flutter Demo Bar Chart'),
    );
  }
}

class MyHomePage extends StatefulWidget {
  const MyHomePage({Key? key, required this.title}) : super(key: key);

  final String title;

  @override
  State<MyHomePage> createState() => _MyHomePageState();
}

class _MyHomePageState extends State<MyHomePage> {
  @override
  Widget build(BuildContext context) {
    return Scaffold(
      appBar: AppBar(
        title: Text(widget.title),
      ),
      body: Padding(
        padding: const EdgeInsets.all(16.0),
        child: Center(
            child: BarChart(BarChartData(
                titlesData: FlTitlesData(
                    show: true,
                    rightTitles: SideTitles(showTitles: false),
                    // leftTitles: SideTitles(
                    //     showTitles: true,
                    //     getTitles: (value) {
                    //       return value.toString();
                    //     }),
                    topTitles: SideTitles(showTitles: false)),
                borderData: FlBorderData(
                    border: const Border(
                  top: BorderSide.none,
                  right: BorderSide.none,
                  left: BorderSide(width: 1),
                  bottom: BorderSide(width: 1),
                )),
                barGroups: [
              BarChartGroupData(x: 1, barRods: [
                BarChartRodData(y: 1.26, width: 15, colors: [Colors.amber]),
              ]),
              BarChartGroupData(x: 2, barRods: [
[...]

PR #833, opened 2021-12-01
Fix interval below one, #811

fatibahrami771998:
Fixes issue #811


imaNNeoFighT:
Thanks for contributing!

PR #835, opened 2021-12-02
fix horizontal over position

666nana666:
fix over position in top and bottom titles


imaNNeoFighT:
Please tell me more about it.

666nana666:
sorry just replied

i will give result about it 

**Before**
image

**After**
image

in this case i need value highest and value lowest but if the value at first value or last value the titles verticals will over position like pic number one so i fixed 

sorry my english is bad hope you undrastand

imaNNeoFighT:
Can you give me your reproducible code?

666nana666:
sure,but i need time.

PR #854, opened 2021-12-13
Make reserved space act like reserved space and not maximum space

kzlakowski:
Hi, it's me again tackling titles issues. The TextAlign PR I made doesn't  work as well as I expected because the texts don't have place to be aligned into. The `SideTitles` class has a field called `reservedSize` but it only sets texts max width. This PR fixes that.
Here are some examples of the issue: 

Aligned to left:
Before:
Screenshot_1639396244
After:
Screenshot_1639396236
Aligned to center:
Before:
Screenshot_1639396253
After:
Screenshot_1639396257

### DIFF
--- a/lib/src/chart/bar_chart/bar_chart_painter.dart
+++ b/lib/src/chart/bar_chart/bar_chart_painter.dart
@@ -408,7 +408,10 @@ class BarChartPainter extends AxisChartPainter<BarChartData> {
               textDirection: leftTitles.textDirection,
               textScaleFactor: holder.textScale,
             );
-            tp.layout(maxWidth: leftTitles.reservedSize);
+            tp.layout(
+              maxWidth: leftTitles.reservedSize,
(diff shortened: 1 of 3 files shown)


imaNNeoFighT:
Good catch!
Can you please update wherever we have similar code please?
(You can search and find all of them).

kzlakowski:
> Good catch! Can you please update wherever we have similar code? (You can search and find all of them).

I believe I have done so, as I understand only Line,Bar and Scatter charts use Titles in this fashion and I've updated all 3 of them. I deliberately skipped top and bottom tiles since they haven't been using their reserved space to constrain their widths. If any code parts that might require update come to your mind, please let me know and I will look into it.

imaNNeoFighT:
Your PR's are so straightforward.
Good description and great implementation. Robust and clean.
Thanks for contributing!
Hope to see more from you!

PR #891, opened 2022-02-03
Add option to include y distance for nearest touched spot

### LINKED ISSUE #716: LineChart Is it possible to show tooltips and dots when the users only touch the line
Hi, im currently using LineChart for our company's project. I have problem implementing according to my ui & ux design. Default behaviour is it will show tooltips and dot when we touch every area inside the chart. Is it possible to show tooltips and dots when the users touch only for the line? thank you.

joeldomke:
Instead of only considering the distance on the x axis when checking for the closest spot, this feature allows to choose the closest spot based on Euclidean distance

TODO: 
- [x] Write tests

Fixes #716, #261

### DIFF
--- a/lib/src/chart/line_chart/line_chart_data.dart
+++ b/lib/src/chart/line_chart/line_chart_data.dart
@@ -1397,6 +1397,9 @@ class LineTouchData extends FlTouchData<LineTouchResponse> with EquatableMixin {
   /// Distance threshold to handle the touch event.
   final double touchSpotThreshold;
 
+  /// Distance function used when finding closest points to touch point
+  final CalculateTouchDistance distanceCalculator;
+
   /// Determines to handle default built-in touch responses,
   /// [LineTouchResponse] shows a tooltip popup above the touched spot.
   final bool handleBuiltInTouches;
@@ -1433,6 +1436,7 @@ class LineTouchData extends FlTouchData<LineTouchResponse> with EquatableMixin {
     LineTouchTooltipData? touchTooltipData,
     GetTouchedSpotIndicator? getTouchedSpotIndicator,
     double? touchSpotThreshold,
+    CalculateTouchDistance? distanceCalculator,
     bool? handleBuiltInTouches,
     GetTouchLineY? getTouchLineStart,
     GetTouchLineY? getTouchLineEnd,
@@ -1440,6 +1444,7 @@ class LineTouchData extends FlTouchData<LineTouchResponse> with EquatableMixin {
         getTouchedSpotIndicator =
             getTouchedSpotIndicator ?? defaultTouchedIndicators,
         touchSpotThreshold = touchSpotThreshold ?? 10,
+        distanceCalculator = distanceCalculator ?? xDistance,
         handleBuiltInTouches = handleBuiltInTouches ?? true,
         getTouchLineStart = getTouchLineStart ?? defaultGetTouchLineStart,
         getTouchLineEnd = getTouchLineEnd ?? defaultGetTouchLineEnd,
@@ -1454,6 +1459,7 @@ class LineTouchData extends FlTouchData<LineTouchResponse> with EquatableMixin {
     LineTouchTooltipData? touchTooltipData,
     GetTouchedSpotIndicator? getTouchedSpotIndicator,
     double? touchSpotThreshold,
+    CalculateTouchDistance? distanceCalculator,
     GetTouchLineY? getTouchLineStart,
     GetTouchLineY? getTouchLineEnd,
     bool? handleBuiltInTouches,
@@ -1481,6 +1487,7 @@ class LineTouchData extends FlTouchData<LineTouchResponse> with EquatableMixin {
         touchTooltipData,
         getTouchedSpotIndicator,
         touchSpotThreshold,
+        distanceCalculator,
         handleBuiltInTouches,
         getTouchLineStart,
         getTouchLineEnd,
@@ -1499,6 +1506,15 @@ typedef GetTouchedSpotIndicator = List<TouchedSpotIndicatorData?> Function(
 /// Used for determine the touch indicator line's starting/end point.
 typedef GetTouchLineY = double Function(
     LineChartBarData barData, int spotIndex);
+
+/// Used to calculate the distance between coordinates of a touch event and a spot
+typedef CalculateTouchDistance = double Function(
+    Offset touchPoint, Offset spotPixelCoordinates);
+
+/// Default distanceCalculator only considers distance on x axis
+double xDistance(Offset touchPoint, Offset spotPixelCoordinates) {
+  return ((touchPoint.dx - spotPixelCoordinates.dx)).abs();
+}
 
 /// Default presentation of touched indicators.
 List<TouchedSpotIndicatorData> defaultTouchedIndicators(
@@ -1675,6 +1691,19 @@ class LineBarSpot extends FlSpot with EquatableMixin {
       ];
 }
 
+/// A [LineBarSpot] that holds information about the event that selected it
+class TouchLineBarSpot extends LineBarSpot {
+  /// Distance in pixels from where the user taped
+  final double distance;
+
+  TouchLineBarSpot(
+    LineChartBarData bar,
+    int barIndex,
+    FlSpot spot,
+    this.distance,
+  ) : super(bar, barIndex, spot);
+}
+
 /// Holds data of showing each row item in the tooltip popup.
 class LineTooltipItem with EquatableMixin {
   /// Showing text.
@@ -1760,16 +1789,17 @@ class ShowingTooltipIndicators with EquatableMixin {
 class LineTouchResponse extends BaseTouchResponse {
   /// touch happened on these spots
   /// (if a single line provided on the chart, [lineBarSpots]'s length will be 1 always)
-  final List<LineBarSpot>? lineBarSpots;
+  final List<TouchLineBarSpot>? lineBarSpots;
 
   /// If touch happens, [LineChart] processes it internally and
   /// passes out a list of [lineBarSpots] it gives you information about the touched spot.
+  /// They are sorted based on their distance to the touch event
   LineTouchResponse(this.lineBarSpots) : super();
 
   /// Copies current [LineTouchResponse] to a new [LineTouchResponse],
   /// and replaces provided values.
   LineTouchResponse copyWith({
-    List<LineBarSpot>? lineBarSpots,
+    List<TouchLineBarSpot>? lineBarSpots,
   }) {
     return LineTouchResponse(
       lineBarSpots ?? this.lineBarSpots,
--- a/lib/src/chart/line_chart/line_chart_data.dart
+++ b/lib/src/chart/line_chart/line_chart_data.dart
@@ -1444,7 +1444,7 @@ class LineTouchData extends FlTouchData<LineTouchResponse> with EquatableMixin {
         getTouchedSpotIndicator =
             getTouchedSpotIndicator ?? defaultTouchedIndicators,
         touchSpotThreshold = touchSpotThreshold ?? 10,
-        distanceCalculator = distanceCalculator ?? xDistance,
+        distanceCalculator = distanceCalculator ?? _xDistance,
         handleBuiltInTouches = handleBuiltInTouches ?? true,
         getTouchLineStart = getTouchLineStart ?? defaultGetTouchLineStart,
         getTouchLineEnd = getTouchLineEnd ?? defaultGetTouchLineEnd,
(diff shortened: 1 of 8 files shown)


imaNNeoFighT:
#892

imaNNeoFighT:
I will check it soon.

joeldomke:
Just to be clear, this PR and #892 to are independent of each other, although it might make sense to use them together in many cases.

joeldomke:
Initially I liked your approach since it fixes #892 at the same time. But it actually misses the case mentioned in #892 where two points of two lines have the same x and y coordinate. 

Adding the distance property mentioned in #892 might also be useful for the case where we only consider the x distance.
Let's consider the following case where we have one line with a data point at x = 5 and another line with a data point at x = 6. Now the user taps on x = 5.2. Both data points would be returned in the `LineTouchResponse` which will lead to the following selection. I don't think there currently is a way to prevent this.

Screenshot_1643974773

imaNNeoFighT:
There is a `touchSpotThreshold` property in the [LineTouchData](https://github.com/imaNNeoFighT/fl_chart/blob/master/repo_files/documentations/line_chart.md#linetouchdata-read-about-touch-handling) class. 
The default value of `touchSpotThreshold` is `10`, leading to this problem. You can try a lower value like `3` to fix the mentioned problem.

joeldomke:
Lowering the threshold does not really fix the problem. After lowering it (which I maybe don't want because I want certain user interaction) I could just move the two points close together and would again face the same issue.

joeldomke:
I switched to a new approach where the user can set a distance function. Also I added the distance to `LineBarSpot`. I still have to clean up and format the code.

lib/src/chart/line_chart/line_chart_data.dart
```diff
@@ -11,7 +11,7 @@ import 'package:flutter/material.dart' hide Image;
 /// [LineChart] needs this class to render itself.
 ///
 /// It holds data needed to draw a line chart,
-/// including bar lines, spots, colors, touches, ...
+/// incluing bar lines, spots, colors, touches, ...
```
imaNNeoFighT, line 14: I think you removed a character here :)

lib/src/chart/line_chart/line_chart_data.dart
```diff
@@ -1397,6 +1397,10 @@ class LineTouchData extends FlTouchData<LineTouchResponse> with EquatableMixin {
   /// Distance threshold to handle the touch event.
   final double touchSpotThreshold;
 
+  /// Distance function used when finding closest points to touch point
+  final double Function(Offset touchPoint, Offset spotPixelCoordinates)
```
imaNNeoFighT, line 1401: Please make a typedef for this function definition.
Something like this:
```dart
/// Bla bala bla bla.
typedef CalculateTouchDistance = double Function(Offset touchPoint, Offset spotPixelCoordinates);
 ```

File: lib/src/chart/line_chart/line_chart_data.dart [hunk 3]
```diff
@@ -1397,6 +1397,10 @@ class LineTouchData extends FlTouchData<LineTouchResponse> with EquatableMixin {
...
+      distanceFunction;
```
imaNNeoFighT, line 1402: I think `distanceCalculator` is a better name for this.

lib/src/chart/line_chart/line_chart_data.dart
```diff
@@ -1433,13 +1437,16 @@ class LineTouchData extends FlTouchData<LineTouchResponse> with EquatableMixin {
     LineTouchTooltipData? touchTooltipData,
     GetTouchedSpotIndicator? getTouchedSpotIndicator,
     double? touchSpotThreshold,
+    double Function(Offset, Offset)? distanceFunction,
```
imaNNeoFighT, line 1440: Use the defined typedef here.

lib/src/chart/line_chart/line_chart_data.dart
```diff
@@ -1433,13 +1437,16 @@ class LineTouchData extends FlTouchData<LineTouchResponse> with EquatableMixin {
...
     bool? handleBuiltInTouches,
     GetTouchLineY? getTouchLineStart,
     GetTouchLineY? getTouchLineEnd,
   })  : touchTooltipData = touchTooltipData ?? LineTouchTooltipData(),
         getTouchedSpotIndicator =
             getTouchedSpotIndicator ?? defaultTouchedIndicators,
         touchSpotThreshold = touchSpotThreshold ?? 10,
+        distanceFunction = distanceFunction ??
```
imaNNeoFighT, line 1448: Please add this property in `copyWith` function too.

lib/src/chart/line_chart/line_chart_data.dart
```diff
@@ -1653,6 +1661,9 @@ class LineBarSpot extends FlSpot with EquatableMixin {
   /// Is the index of our [super.spot], in the [LineChartBarData.spots] list.
   final int spotIndex;
 
+  /// Distance in pixels from where the user taped
+  final double distance;
```
imaNNeoFighT, line 1665: Why do we need this?
`LineBarSpot` is a general class. It is not only for touched spots. It's a wrapper to hold a spot location.

joeldomke: Hmm, yes you are right. My intention was to give more control to the user, who then can choose which spots to select. But maybe for now I should just go with your suggestion, to return the LineBarSpots as they are returned currently, along with the LineBarSpot that is the closest. (As described here: https://github.com/imaNNeoFighT/fl_chart/pull/891#issuecomment-1029808580)

Let me take a look at that.

lib/src/chart/line_chart/line_chart_painter.dart
```diff
@@ -1521,17 +1521,20 @@ class LineChartPainter extends AxisChartPainter<LineChartData> {
       }
     }
 
+    touchedSpots.sort((a, b) => a.distance.compareTo(b.distance));
```
joeldomke, line 1524: I don't think changing the order of 'touchedSpots' is a problem here?

imaNNeoFighT: I don't think too. It looks good to me.

lib/src/chart/line_chart/line_chart_renderer.dart
```diff
@@ -90,11 +90,11 @@ class RenderLineChart extends RenderBaseChart<LineTouchResponse> {
...
   @override
   LineTouchResponse getResponseAtLocation(Offset localPosition) {
-    var touchedSpots = painter.handleTouch(
+    final touchedSpots = painter.handleTouch(
       localPosition,
       mockTestSize ?? size,
       paintHolder,
     );
-    return LineTouchResponse(touchedSpots);
+    return LineTouchResponse(touchedSpots, touchedSpots?.first);
```
joeldomke, line 98: I guess we could keep `LineTouchResponse` at it was before, only containing the list of touched spots and just document the fact, that they are sorted by distance? There isn't really any need to save the list along with the first element of the list, right?

I think that's the final decision I need your thoughts on. Then I would write the tests, after which I would be finished with coding.

imaNNeoFighT: I agree with removing it. Go ahead and when everything was ready let me review everything again.

imaNNeoFighT:
Any news on this?

joeldomke:
Should be finished now

lib/src/chart/line_chart/line_chart_data.dart
```diff
@@ -1764,12 +1793,14 @@ class LineTouchResponse extends BaseTouchResponse {
...
   /// If touch happens, [LineChart] processes it internally and
   /// passes out a list of [lineBarSpots] it gives you information about the touched spot.
+  /// They are sorted based on their distance to the touch event
   LineTouchResponse(this.lineBarSpots) : super();
 
   /// Copies current [LineTouchResponse] to a new [LineTouchResponse],
   /// and replaces provided values.
   LineTouchResponse copyWith({
     List<LineBarSpot>? lineBarSpots,
+    LineBarSpot? closestLineBarSpot,
```
imaNNeoFighT, line 1803: I think you forgot to remove this line

lib/src/chart/line_chart/line_chart_painter.dart
```diff
@@ -1507,7 +1507,7 @@ class LineChartPainter extends AxisChartPainter<LineChartData> {
 
     /// it holds list of nearest touched spots of each line
     /// and we use it to draw touch stuff on them
-    final touchedSpots = <LineBarSpot>[];
+    final touchedSpots = <TouchLineBarSpot>[];
```
imaNNeoFighT, line 1510: We need to update the return type of this function from `List<LineBarSpot>` to `List<TouchLineBarSpot>`.

It's a breaking change and we need to notify our users.

imaNNeoFighT: It would be great to add these changes to our [CHANGELOG.md](https://github.com/imaNNeoFighT/fl_chart/blob/dev/CHANGELOG.md).

I should update our [CONTRIBUTING.md](https://github.com/imaNNeoFighT/fl_chart/blob/master/CONTRIBUTING.md) guideline to explain how CHANGELOG.md works. But it's straightforward. You can check our previous versions changelog.

imaNNeoFighT:
I wrote my last comments. Also, you must rebase this branch with our `dev` branch. (It seems there are some minor conflicts)

lib/src/chart/line_chart/line_chart_painter.dart
joeldomke: I added changed the type in `LineTouchResponse` and added it to the changelog

joeldomke:
I merged dev into my branch, did this solve the conflicts?

imaNNeoFighT:
No, it doesn't solve the problem.
You need to rebase with the dev branch.
I think you can reset your commits and just make one commit containing your needed changes. (Now it's 11 commits)

joeldomke:
I squashed them into one and rebased with the dev branch

lib/src/chart/line_chart/line_chart_data.dart
```diff
@@ -1454,6 +1459,7 @@ class LineTouchData extends FlTouchData<LineTouchResponse> with EquatableMixin {
     LineTouchTooltipData? touchTooltipData,
     GetTouchedSpotIndicator? getTouchedSpotIndicator,
     double? touchSpotThreshold,
+    CalculateTouchDistance? distanceCalculator,
```
imaNNeoFighT, line 1462: You added this input parameter in `copyWith` function. But you didn't use it in the body.

lib/src/chart/line_chart/line_chart_data.dart
```diff
@@ -1500,6 +1507,15 @@ typedef GetTouchedSpotIndicator = List<TouchedSpotIndicatorData?> Function(
 typedef GetTouchLineY = double Function(
     LineChartBarData barData, int spotIndex);
 
+/// Used to calculate the distance between coordinates of a touch event and a spot
+typedef CalculateTouchDistance = double Function(
+    Offset touchPoint, Offset spotPixelCoordinates);
+
+/// Default distanceCalculator only considers distance on x axis
+double xDistance(Offset touchPoint, Offset spotPixelCoordinates) {
```
imaNNeoFighT, line 1515: It can be private.

lib/src/chart/line_chart/line_chart_data.dart
```diff
@@ -1675,6 +1691,19 @@ class LineBarSpot extends FlSpot with EquatableMixin {
       ];
 }
 
+/// A [LineBarSpot] that holds information about the event that selected it
+class TouchLineBarSpot extends LineBarSpot {
```
imaNNeoFighT, line 1695: We document our lineChart data classes in [line_chart.md](https://github.com/imaNNeoFighT/fl_chart/blob/master/repo_files/documentations/line_chart.md) in a table. 
Please don't forget to add this class in line_chart.md (with the correct order, just like other data classes).

imaNNeoFighT:
And about the CHANGELOG.md, thank you for adding BREAKING change.
But also we have an **IMPROVEMENT**. 
You made the touch logic customizable!
We need to let users know how they can customize the touch logic.

imaNNeoFighT:
Any update? 
We need to release a version, and I'm waiting for this branch.
Do you have any estimation on this?

lib/src/chart/line_chart/line_chart_data.dart
joeldomke: I added documentation, but I am not sure about the default values. `TouchLineBarSpot` does not have default values, but it's the same for `LineBarSpot` and there it currently says `null` as default, so I did the same.

joeldomke:
Yes, it's hopefully done now. And also thank your for your patience. This is my first PR and there were a lot of things I missed at first, but I enjoyed it.

imaNNeoFighT:
Thank you @joeldomke 
Don't worry, that was good!
It has just merged into the `dev` branch.
I'm going to bump up a version :)

It will be great if you modify (or add) a chart to showcase this feature.
I hope to see more from you :))

Issue #897, opened 2022-02-10
Feature to clip the data point along with radius.


prajwal27:
**Describe the bug**
The Scatter chart in the image below does not clip the bubble.It considers only the data points. The `ScatterChartData` has a attribute called `clipData` , which checks if the data point is within the chart and then decides to clip it or not. 
However, it doesn't consider the radius of the bubble and hence, even if the data point is inside, the bubble might be so big that it causes it to go outside the axis, which is not a good UI/UX.

**To Reproduce**
It is fairly easy to reproduce. Take any example of scatter chart and increase the radius until you see it cross the axis.

**Screenshots**
![image](https://user-images.githubusercontent.com/95603936/153347535-e485e394-6eee-44cd-955e-ccc49631817b.png)

**Possible Solution**
So, currently we have clipped the bubble (see the topmost bubble), and it doesn't cross the chart, but it does cross the axis. So, what I feel would be a better idea is to crop clip the bubble inside the chart. So, we get something like this
![image](https://user-images.githubusercontent.com/95603936/153347359-37148589-2c5b-4216-b678-ebe236817f62.png)
Here, the bubble doesn't cross the axis.

**Versions**
 - Flutter version: 2.10.0
 - fl_chart version: 0.41.0

imaNNeoFighT:
It has been fixed in [0.45.0](https://github.com/imaNNeoFighT/fl_chart/releases/tag/0.45.0) with your help :))
Thank you!

PR #898, opened 2022-02-10
Implement `clipData` feature in scatter chart.

prajwal27:
#897.

The Scatter chart can be used as a Bubble chart by providing different radii. So, when the radius is bigger and close to the axis of the chart, it is used to cross the axis and go out of the chart.

So, in my approach, I've added a new attribute in `ScatterChartData` called `clipBubble`. If we set it to true, it will clip the chart before drawing the spots such that the bubbles never cross the chart's axes. After the bubbles are drawn, it again restores the original region.

### DIFF
--- a/lib/src/chart/scatter_chart/scatter_chart_data.dart
+++ b/lib/src/chart/scatter_chart/scatter_chart_data.dart
@@ -3,9 +3,9 @@ import 'dart:ui';
 
 import 'package:equatable/equatable.dart';
 import 'package:fl_chart/fl_chart.dart';
+import 'package:fl_chart/src/extensions/color_extension.dart';
 import 'package:fl_chart/src/utils/lerp.dart';
 import 'package:flutter/material.dart';
-import 'package:fl_chart/src/extensions/color_extension.dart';
 
 import 'scatter_chart_helper.dart';
 
@@ -18,6 +18,7 @@ class ScatterChartData extends AxisChartData with EquatableMixin {
   final FlTitlesData titlesData;
   final ScatterTouchData scatterTouchData;
   final List<int> showingTooltipIndicators;
+  final bool clipBubble;
 
   /// [ScatterChart] draws some points in a square space,
   /// points are defined by [scatterSpots],
@@ -39,6 +40,9 @@ class ScatterChartData extends AxisChartData with EquatableMixin {
   /// just put spot indices you want to show it on top of them.
   ///
   /// [clipData] forces the [LineChart] to draw lines inside the chart bounding box.
+  ///
+  /// [clipBubble] forces the [ScatterChart] to crop the bubbles and make sure it does
+  /// not cross the chart bounding box.
   ScatterChartData({
     List<ScatterSpot>? scatterSpots,
     FlTitlesData? titlesData,
@@ -53,10 +57,12 @@ class ScatterChartData extends AxisChartData with EquatableMixin {
     double? maxY,
     FlClipData? clipData,
     Color? backgroundColor,
+    bool? clipBubble,
   })  : scatterSpots = scatterSpots ?? const [],
         titlesData = titlesData ?? FlTitlesData(),
         scatterTouchData = scatterTouchData ?? ScatterTouchData(),
         showingTooltipIndicators = showingTooltipIndicators ?? const [],
+        clipBubble = clipBubble ?? false,
         super(
           gridData: gridData ?? FlGridData(),
           touchData: scatterTouchData ?? ScatterTouchData(),
@@ -101,6 +107,7 @@ class ScatterChartData extends AxisChartData with EquatableMixin {
         minY: lerpDouble(a.minY, b.minY, t),
         maxY: lerpDouble(a.maxY, b.maxY, t),
         clipData: b.clipData,
+        clipBubble: b.clipBubble,
         backgroundColor: Color.lerp(a.backgroundColor, b.backgroundColor, t),
       );
     } else {
@@ -124,6 +131,7 @@ class ScatterChartData extends AxisChartData with EquatableMixin {
     double? maxY,
     FlClipData? clipData,
     Color? backgroundColor,
+    bool? clipBubble,
   }) {
     return ScatterChartData(
       scatterSpots: scatterSpots ?? this.scatterSpots,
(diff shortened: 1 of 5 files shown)


imaNNeoFighT:
@prajwal27 Thanks for contributing!

`clipData` property (defined in `AxisChartData`, which `ScatterChartData` extends from) is responsible to do exactly what you are looking for. I checked our `scatter_chart_painter.dart`; it seems we forgot to use it. It has no implementation logic at the moment.

We need to fix the functionality of `clipData` property to fix your issue.
Cheers!

imaNNeoFighT:
You can check [line_chart_painter.dart:69](https://github.com/imaNNeoFighT/fl_chart/blob/860a00b5d6d9b4c11864775882d7987f8d98a48a/lib/src/chart/line_chart/line_chart_painter.dart#L69) to get the idea.

prajwal27:
So @imaNNeoFighT , shall I remove the new `clipBubble` attribute ?

imaNNeoFighT:
Yes, you're correct!
We need to use `clipData` to do that.

prajwal27:
@imaNNeoFighT  I checked the line_chart_painter.dart and I got this 
```
 if (data.clipData.any) {
      canvasWrapper.saveLayer(
        Rect.fromLTWH(0, -40, canvasWrapper.size.width + 40,
            canvasWrapper.size.height + 40),
        Paint(),
      );
```

What's the logic of 40 here?

imaNNeoFighT:
To be honest I don't remember :))))
Try to remove it in the scatter chart and see does it work or not.

prajwal27:
@imaNNeoFighT  I've updated the PR as you've suggested. I did try and remove the 40. I didn't notice any change. 
Let me know if this is good to merge.
PS: Our team is relying on this fix :) It'll be good if it can be merged asap even is the package gets upgraded later.

lib/src/chart/scatter_chart/scatter_chart_data.dart
```diff
@@ -3,9 +3,9 @@ import 'dart:ui';
 
 import 'package:equatable/equatable.dart';
 import 'package:fl_chart/fl_chart.dart';
+import 'package:fl_chart/src/extensions/color_extension.dart';
```
imaNNeoFighT, line 6: This change is unnecessary (It was just removed from line 8 and added in line 6). Please revert all changes like this.

imaNNeoFighT:
Hi @prajwal27.
It's good to merge. Also, it would be great if you update the [CHANGELOG.md](https://github.com/imaNNeoFighT/fl_chart/blob/dev/CHANGELOG.md) file and add your fix.

lib/src/chart/scatter_chart/scatter_chart_data.dart
prajwal27: @imaNNeoFighT  I think it was auto formatted. I'll check it again.

prajwal27:
> Hi @prajwal27. It's good to merge. Also, it would be great if you update the [CHANGELOG.md](https://github.com/imaNNeoFighT/fl_chart/blob/dev/CHANGELOG.md) file and add your fix.

Yeah, is this considered as bug fix or enhancement?

imaNNeoFighT:
> > Hi @prajwal27. It's good to merge. Also, it would be great if you update the [CHANGELOG.md](https://github.com/imaNNeoFighT/fl_chart/blob/dev/CHANGELOG.md) file and add your fix.
> 
> Yeah, is this considered as bug fix or enhancement?

I think it is a bug fix (Because it was supposed to work).
Don't forget to mention #897  issue at the end of line (like other changelogs)

lib/src/chart/scatter_chart/scatter_chart_data.dart
imaNNeoFighT: Thank you!

prajwal27: @imaNNeoFighT  I'm certain that it's because of the dart format. Do you still want me to revert it?

imaNNeoFighT: It's better to have features and format changes in different PRs. (Or maybe different commits)

prajwal27: ok. I'll change it back.

imaNNeoFighT: Thank you!

prajwal27: I'm unable to change it, as it gets auto formatted immediately.

prajwal27:
@imaNNeoFighT  thanks for the quick suggestions and reviews. It would be great if you could merge this.

imaNNeoFighT:
Thank you @prajwal27 
I'm trying my best to merge it ASAP.

There is a minor problem:



As you see, it doesn't work correctly with our borders (But we considered the border width in our code).
I checked it. It has a minor problem. 

We should:
Add the left and top `(borderWidth / 2)`
Minus the right and top `(borderWidth / 2)`

I know you just copied it from our LineChart. 
But the LineChart has exact same bug too, I'm going to fix it in another PR :)))

Below code is correct:
```dart
if (clip.left) {
  final borderWidth = border?.left.width ?? 0;
  left = getLeftOffsetDrawSize(holder) + (borderWidth / 2);
}
if (clip.top) {
  final borderWidth = border?.top.width ?? 0;
  top = getTopOffsetDrawSize(holder) + (borderWidth / 2);
}
if (clip.right) {
  final borderWidth = border?.right.width ?? 0;
  right = getLeftOffsetDrawSize(holder) +
      chartUsableSize.width -
      (borderWidth / 2);
}
if (clip.bottom) {
  final borderWidth = border?.bottom.width ?? 0;
  bottom = getTopOffsetDrawSize(holder) +
      chartUsableSize.height -
      (borderWidth / 2);
}
```

prajwal27:
@imaNNeoFighT  that's great. I didn't see it coming. Are we good to merge now?

imaNNeoFighT:
No please solve the problem I mentioned first, then we're going to merge it. (I said I will fix it in the LineChart)

imaNNeoFighT:
Landed in `dev`.
Thanks for contributing!

prajwal27:
It was fun understanding the code! You'll probably see me more often than not :)

imaNNeoFighT:
Thanks for being patient :)))
You know it's frustrating for some people. 
But I need to make sure to merge a high-quality code.

PR #907, opened 2022-02-17
Label feature in Scatter Chart

### LINKED ISSUE #902: Feature to add labels around the spot in scatter chart
I would like to have a feature where the item labels are around the spot in scatter chart. If the raidus is big enough, it can be inside the circle as well.
image

My Solution is to draw the text after drawing the spots. However, the key thing is to position it below/above/left/right of the spot so that it doesn't cross the chart.

**Describe alternatives you've considered**

**Additional context**
I've shared the desired output above.

prajwal27:
fix #902 
1. Allows giving `label` to `ScatterSpot` in Scatter Chart.
2. Draws the label based on the `ScatterLabelSettings` provided.

### DIFF
--- a/lib/src/chart/scatter_chart/scatter_chart_data.dart
+++ b/lib/src/chart/scatter_chart/scatter_chart_data.dart
@@ -187,19 +187,19 @@ class ScatterSpot extends FlSpot with EquatableMixin {
   /// Determines color of the spot.
   Color color;
 
+  /// Determines label of the spot.
+  final String? label;
+
   /// You can change [show] value to show or hide the spot,
   /// [x], and [y] defines the location of spot in the [ScatterChart],
   /// [radius] defines the size of spot, and [color] defines the color of it.
-  ScatterSpot(
-    double x,
-    double y, {
-    bool? show,
-    double? radius,
-    Color? color,
-  })  : show = show ?? true,
+  ScatterSpot(double x, double y,
+      {bool? show, double? radius, Color? color, String? label})
+      : show = show ?? true,
         radius = radius ?? 6,
         color = color ??
             Colors.primaries[((x * y) % Colors.primaries.length).toInt()],
+        label = label,
         super(x, y);
 
   @override
@@ -209,6 +209,7 @@ class ScatterSpot extends FlSpot with EquatableMixin {
     bool? show,
     double? radius,
     Color? color,
+    String? label,
   }) {
     return ScatterSpot(
       x ?? this.x,
@@ -216,6 +217,7 @@ class ScatterSpot extends FlSpot with EquatableMixin {
       show: show ?? this.show,
       radius: radius ?? this.radius,
       color: color ?? this.color,
+      label: label,
     );
   }
 
@@ -227,6 +229,7 @@ class ScatterSpot extends FlSpot with EquatableMixin {
       show: b.show,
       radius: lerpDouble(a.radius, b.radius, t),
       color: Color.lerp(a.color, b.color, t),
+      label: b.label,
     );
   }
 
@@ -238,6 +241,7 @@ class ScatterSpot extends FlSpot with EquatableMixin {
         show,
         radius,
         color,
+        label,
       ];
 }
 
--- a/lib/src/chart/scatter_chart/scatter_chart_data.dart
+++ b/lib/src/chart/scatter_chart/scatter_chart_data.dart
@@ -18,6 +18,7 @@ class ScatterChartData extends AxisChartData with EquatableMixin {
   final FlTitlesData titlesData;
   final ScatterTouchData scatterTouchData;
   final List<int> showingTooltipIndicators;
+  final ScatterLabelSettings scatterLabelSettings;
 
   /// [ScatterChart] draws some points in a square space,
   /// points are defined by [scatterSpots],
@@ -55,10 +56,12 @@ class ScatterChartData extends AxisChartData with EquatableMixin {
     double? baselineY,
     FlClipData? clipData,
     Color? backgroundColor,
+    ScatterLabelSettings? scatterLabelSettings,
   })  : scatterSpots = scatterSpots ?? const [],
         titlesData = titlesData ?? FlTitlesData(),
         scatterTouchData = scatterTouchData ?? ScatterTouchData(),
         showingTooltipIndicators = showingTooltipIndicators ?? const [],
+        scatterLabelSettings = scatterLabelSettings ?? ScatterLabelSettings(),
         super(
           gridData: gridData ?? FlGridData(),
           touchData: scatterTouchData ?? ScatterTouchData(),
@@ -108,6 +111,7 @@ class ScatterChartData extends AxisChartData with EquatableMixin {
         baselineY: lerpDouble(a.baselineY, b.baselineY, t),
         clipData: b.clipData,
         backgroundColor: Color.lerp(a.backgroundColor, b.backgroundColor, t),
+        scatterLabelSettings: b.scatterLabelSettings,
       );
     } else {
       throw Exception('Illegal State');
@@ -132,6 +136,7 @@ class ScatterChartData extends AxisChartData with EquatableMixin {
     double? baselineY,
     FlClipData? clipData,
     Color? backgroundColor,
+    ScatterLabelSettings? scatterLabelSettings,
   }) {
     return ScatterChartData(
       scatterSpots: scatterSpots ?? this.scatterSpots,
@@ -150,6 +155,7 @@ class ScatterChartData extends AxisChartData with EquatableMixin {
       baselineY: baselineY ?? this.baselineY,
       clipData: clipData ?? this.clipData,
       backgroundColor: backgroundColor ?? this.backgroundColor,
+      scatterLabelSettings: scatterLabelSettings ?? this.scatterLabelSettings,
     );
   }
 
@@ -173,6 +179,7 @@ class ScatterChartData extends AxisChartData with EquatableMixin {
         maxY,
         baselineY,
         rangeAnnotations,
+        scatterLabelSettings,
       ];
 }
 
@@ -574,4 +581,50 @@ class ScatterChartDataTween extends Tween<ScatterChartData> {
   ScatterChartData lerp(double t) {
     return begin!.lerp(begin!, end!, t);
   }
+}
+
+/// Defines information about the labels in the [ScatterChart]
+class ScatterLabelSettings with EquatableMixin {
+  /// Determines whether to show or hide the labels
+  final bool showLabel;
+
+  /// Determines style of the labels
+  final TextStyle? textStyle;
+
+  /// You can change [showLabel] value to show or hide the label,
+  /// [textStyle] defines the style of label in the [ScatterChart].
+  ScatterLabelSettings({
+    bool? showLabel,
+    TextStyle? textStyle,
+  })  : showLabel = showLabel ?? false,
+        textStyle = textStyle;
+
+  ScatterLabelSettings copyWith({
+    bool? showLabel,
+    TextStyle? textStyle,
+  }) {
+    return ScatterLabelSettings(
+      showLabel: showLabel ?? this.showLabel,
+      textStyle: textStyle,
+    );
+  }
+
+  /// Lerps a [ScatterLabelSettings] based on [t] value, check [Tween.lerp].
+  static ScatterLabelSettings lerp(
+    ScatterLabelSettings a,
+    ScatterLabelSettings b,
+    double t,
+  ) {
+    return ScatterLabelSettings(
+      showLabel: b.showLabel,
+      textStyle: b.textStyle,
+    );
+  }
+
+  /// Used for equality check, see [EquatableMixin].
+  @override
+  List<Object?> get props => [
+        showLabel,
+        textStyle,
+      ];
 }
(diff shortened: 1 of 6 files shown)


prajwal27:
@imaNNeoFighT  Can you check if this feature is good? I'm aware that the tests are yet to be written for this. How do I check locally the tests code coverage?

prajwal27:
I got to know how to check code coverage. It's inbuilt in IDE lol

prajwal27:
@imaNNeoFighT  I've added all the necessary tests. I hope it's good enough to merge. It's a basic feature for now. As and when required, we can try and add more to it.

lib/src/chart/scatter_chart/scatter_chart_data.dart
```diff
@@ -187,19 +194,19 @@ class ScatterSpot extends FlSpot with EquatableMixin {
   /// Determines color of the spot.
   Color color;
 
+  /// Determines label of the spot.
+  final String label;
```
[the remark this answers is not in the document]
imaNNeoFighT: In this way, `ScatterLabelSettings` has a single responsibility (showing titles).

imaNNeoFighT:
> got to know how to check code coverage. It's in

You can also use provided command in our [Makefile](https://github.com/imaNNeoFighT/fl_chart/blob/master/Makefile) using `make showTestCoverage`.

lib/src/chart/scatter_chart/scatter_chart_data.dart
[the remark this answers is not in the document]
prajwal27: @imaNNeoFighT  Possible. Also, we should give the flexibility to have a different text style for each of the labels. We shouldn't force a single text style for all the labels.

prajwal27:
@imaNNeoFighT  I've modified the code now. Let me know if it's good.

prajwal27:
@imaNNeoFighT  any update on the PR ?

lib/src/chart/scatter_chart/scatter_chart_data.dart
```diff
@@ -108,6 +111,7 @@ class ScatterChartData extends AxisChartData with EquatableMixin {
         baselineY: lerpDouble(a.baselineY, b.baselineY, t),
         clipData: b.clipData,
         backgroundColor: Color.lerp(a.backgroundColor, b.backgroundColor, t),
+        scatterLabelSettings: b.scatterLabelSettings,
```
imaNNeoFighT, line 114: Here you should use:
```dart
scatterLabelSettings: ScatterLabelSettings.lerp(a.scatterLabelSettings, b.scatterLabelSettings),
```

lib/src/chart/scatter_chart/scatter_chart_data.dart
```diff
@@ -571,3 +578,76 @@ class ScatterChartDataTween extends Tween<ScatterChartData> {
     return begin!.lerp(begin!, end!, t);
   }
 }
+
+/// It gives you the index value of the spot and gets the text style of the label.
+typedef GetLabelTextStyleFunction = TextStyle Function(int spotIndex);
```
imaNNeoFighT, line 583: It's better to pass the spot object here (along with the index)

lib/src/chart/scatter_chart/scatter_chart_data.dart
```diff
@@ -571,3 +578,76 @@ class ScatterChartDataTween extends Tween<ScatterChartData> {
...
+
+/// It gives you the index value of the spot and returns the label of the spot.
+typedef GetLabelFunction = String Function(int spotIndex);
```
imaNNeoFighT, line 586: It's better to pass the spot object here (along with the index)

lib/src/chart/scatter_chart/scatter_chart_painter.dart
```diff
@@ -290,6 +290,67 @@ class ScatterChartPainter extends AxisChartPainter<ScatterChartData> {
       );
     }
 
+    if (data.scatterLabelSettings.showLabel) {
+      for (final scatterSpot in data.scatterSpots) {
+        int spotIndex = data.scatterSpots.indexOf(scatterSpot);
```
imaNNeoFighT, line 295: `indexOf()` call takes O(N) to find the index. While we can use an indexed `for` to have both `index` and `scatterSpot` together.

like:
```dart
for (int i = 0; i < data.scatterSpots.length; i++) {
   scatterSpot = data.scatterSpots[i]
   int spotIndex = i;
}
```

lib/src/chart/scatter_chart/scatter_chart_painter.dart
```diff
@@ -290,6 +290,67 @@ class ScatterChartPainter extends AxisChartPainter<ScatterChartData> {
...
+
+        String label = data.scatterLabelSettings.getLabelFunction(spotIndex);
+
+        if (label.isEmpty || !scatterSpot.show) {
+          continue;
+        }
+
+        final span = TextSpan(
+          text: label,
+          style: data.scatterLabelSettings.getLabelTextStyleFunction(spotIndex),
```
imaNNeoFighT, line 305: We have a `Utils().getThemeAwareTextStyle()` function which helps us to support the app theme by default.
Please use it (check [here](https://github.com/imaNNeoFighT/fl_chart/blob/master/lib/src/chart/line_chart/line_chart_painter.dart#L917) for more info)

lib/src/chart/scatter_chart/scatter_chart_painter.dart
```diff
@@ -290,6 +290,67 @@ class ScatterChartPainter extends AxisChartPainter<ScatterChartData> {
...
+        );
+
+        final tp = TextPainter(
+          text: span,
+          textAlign: TextAlign.center,
+          textDirection: TextDirection.ltr,
```
imaNNeoFighT, line 311: It's better not to hardcode the `textDirection` property because some users have RTL-based language.
You can put a single property in `ScatterLabelSettings` to hold a dynamic value of `textDirection` (with `ltr` as default)

lib/src/chart/scatter_chart/scatter_chart_data.dart
```diff
@@ -571,3 +578,76 @@ class ScatterChartDataTween extends Tween<ScatterChartData> {
...
+
+/// It gives you the default text style of the label for a spot.
+TextStyle getDefaultLabelTextStyleFunction(int spotIndex) {
+  return const TextStyle();
+}
+
+/// It gives you the default label of the spot.
+String getDefaultLabelFunction(int spotIndex) {
+  return '';
```
imaNNeoFighT, line 595: It's better to show a value (for example radius of a spot) by default.

lib/src/chart/scatter_chart/scatter_chart_data.dart
```diff
@@ -571,3 +578,76 @@ class ScatterChartDataTween extends Tween<ScatterChartData> {
...
+
+/// It gives you the index value of the spot and gets the text style of the label.
+typedef GetLabelTextStyleFunction = TextStyle Function(int spotIndex);
+
+/// It gives you the index value of the spot and returns the label of the spot.
+typedef GetLabelFunction = String Function(int spotIndex);
+
+/// It gives you the default text style of the label for a spot.
+TextStyle getDefaultLabelTextStyleFunction(int spotIndex) {
+  return const TextStyle();
```
imaNNeoFighT, line 590: You can make it nullable and return `null` here.
Then we use a textStyle from the parent theme.

prajwal27: > It doesn't work good on our ScatterChartSample2 with this code:
> 
> ```dart
> scatterLabelSettings: ScatterLabelSettings(
>   showLabel: true,
>   getLabelFunction: (index) => index.toString(),
>   getLabelTextStyleFunction: (index) => const TextStyle(
>     color: Colors.white,
>     fontSize: 40,
>   ),
> ),
> ```
> 
> image

@imaNNeoFighT  I think this is because the size of the label is not getting adjusted to the window size. I believe so. What do you think?

prajwal27: it needs to be scaled with the canvas size or the window size. I saw this issue with the axis titles as well, but it is outside the chart. So it doesn't affect much.

prajwal27:
> It doesn't work good on our ScatterChartSample2 with this code:
> 
> ```dart
> scatterLabelSettings: ScatterLabelSettings(
>   showLabel: true,
>   getLabelFunction: (index) => index.toString(),
>   getLabelTextStyleFunction: (index) => const TextStyle(
>     color: Colors.white,
>     fontSize: 40,
>   ),
> ),
> ```
> 
> image

@imaNNeoFighT I think this is because the size of the label is not getting adjusted to the window size. I believe so. What do you think?

imaNNeoFighT:
> @imaNNeoFighT I think this is because the size of the label is not getting adjusted to the window size. I believe so. What do you think?
> 
> it needs to be scaled with the canvas size or the window size. I saw this issue with the axis titles as well, but it is outside the chart. So it doesn't affect much.

I think it's because of calculations. 
In your logic, texts should be center or top/bottom of each scatter spot (based on the spots size and text size)
Correct?

lib/src/chart/scatter_chart/scatter_chart_painter.dart
imaNNeoFighT: Yes, just like other places that we passed context to our methods.

prajwal27:
> > @imaNNeoFighT I think this is because the size of the label is not getting adjusted to the window size. I believe so. What do you think?
> > it needs to be scaled with the canvas size or the window size. I saw this issue with the axis titles as well, but it is outside the chart. So it doesn't affect much.
> 
> I think it's because of calculations. In your logic, texts should be center or top/bottom of each scatter spot (based on the spots size and text size) Correct?

@imaNNeoFighT  yes. I believe that's a good starting solution to having labels.

imaNNeoFighT:
It seems implemented logic has a problem that doesn't work with large fonts.
Can you please check it out?

imaNNeoFighT:
I know, the font size is set dynamically already (even in my sample)
This is not like side titles, here we have size of  drawing font, and we need to place it in the correct location as your logic is supposed to do.

This is your implemented logic:

```dart
/// if the spot is in the lower half of the chart, then draw the label either in the center or above the spot,
/// if the spot is in upper half of the chart, then draw the label either in the center or below the spot.
if (pixelY > centerChartY) {
  /// if either the height or the width of the spot is greater than the radius of the spot, then draw the label above the bubble,
  /// else draw the label inside the bubble.
  var off = (scatterSpot.radius * 1.5 < tp.height ||
    scatterSpot.radius * 1.5 < tp.width)
  ? scatterSpot.radius + tp.height
  : tp.height / 2;

  newPixelY = pixelY - off;
} else {
  /// if either the height or the width of the spot is greater than the radius of the spot, then draw the label below the bubble,
  /// else draw the label inside the bubble.
  var off = (scatterSpot.radius * 1.5 < tp.height ||
    scatterSpot.radius * 1.5 < tp.width)
  ? scatterSpot.radius
  : -tp.height / 2;
  newPixelY = pixelY + off;
}
```

prajwal27:
@imaNNeoFighT  yeah, but the logic is such that it's either inside the bubble (if big enough) or above/below the spot. Are you saying this location solution is incorrect or the implementation in code is incorrect?

prajwal27:
@imaNNeoFighT  In the image you've shared, `0` is below its spot and `2` is below its spot as well. We haven't handled overlapping of labels though.

imaNNeoFighT:
Overlapping is okay (I know it's hard to handle).
`7`, `4`, `3`, and `1` are perfectly located.
See the top space of `0` and `2`. They have different top spaces (It seems `2` is located correctly, but `0` has no top space).
Also, I don't understand `5` and `6`. It seems something is wrong.

image

prajwal27:
Are you running on web? and let me check it once more.

imaNNeoFighT:
I ran it on my android phone. But I think it doesn't matter where you run. It seems there's something wrong with logic.

prajwal27:
@imaNNeoFighT  I'm getting this when I run on Web 
image

imaNNeoFighT:
What if you decrease the size of your web window?

prajwal27:
@imaNNeoFighT  I get like this, 
image

So, it's up to the application/developer to choose the font size based on window screen size. Normally we have font sizes for the texts, which don't alter much with the window size. This label is part of the chart and hence we need to change the size w.r.t the window size.

There's no way of doing this from the package side.

imaNNeoFighT:
Let me check it again.

imaNNeoFighT:
Okay, it seems fine to me. 
I couldn't reproduce that problem again.

imaNNeoFighT:
Thanks for your patience and thanks for contributing!

prajwal27:
@imaNNeoFighT  thanks for considering my PR and merging it!

prajwal27:
@imaNNeoFighT  can you make changes to `CHANGELOG.md` file? I didn't do it :P

imaNNeoFighT:
You're welcome. I'm pretty sure that people are going to use it in their charts.
I hope to see more from you :)

imaNNeoFighT:
> @imaNNeoFighT can you make changes to `CHANGELOG.md` file? I didn't do it :P

Yes, I just did it in [this commit](https://github.com/imaNNeoFighT/fl_chart/commit/bcc1c1cfb19af5e319685543de803add6f05568d) in our `dev` branch. Please check it out.

I updated our [CONTRIBUTING.md](https://github.com/imaNNeoFighT/fl_chart/blob/dev/CONTRIBUTING.md#changelog) last night to let contributors know how to write changelog.

Issue #692, opened 2021-06-03
Border for Tooltip.


P-B1101:
Hello and Thanks for AMAZING package.
It would be great if we could add some border styling for tooltip. until you guys handle tooltip with widget.
Thanks a lot.

emelinepal:
Hello :) For my studies I have to contribute to an opensource project by completing an Issue, can I take this one ?

imaNNeoFighT:
Hi, @emelinepal. 
That would be great. I appreciate it!
You can start by reading our contribution guidelines [here](https://github.com/imaNNeoFighT/fl_chart/blob/master/CONTRIBUTING.md).
Then start working on the code. I will be available if you have any questions.

Thanks!

emelinepal:
Hello @imaNNeoFighT ! I just made a PR, I am waiting for your feedback :)

imaNNeoFighT:
This feature has just landed in [0.50.0](https://github.com/imaNNeoFighT/fl_chart/releases/tag/0.55.0).

Thanks, @emelinepal, for her contribution.

PR #917, opened 2022-02-23
Added Tooltip border color and width support

fpplm:
#692


CHANGELOG.md
```diff
@@ -384,4 +385,4 @@ LineTouchData(
 * restricted to access private classes of the library
 
 
-## 0.0.1 - Released on (2019 June 4)
\ No newline at end of file
+## 0.0.1 - Released on (2019 June 4)
```
imaNNeoFighT, line 388: Why this line is changed?

lib/src/chart/bar_chart/bar_chart_data.dart
```diff
@@ -642,6 +642,12 @@ class BarTouchTooltipData with EquatableMixin {
   /// The tooltip background color.
   final Color tooltipBgColor;
 
+  /// The tooltip border color.
+  final Color tooltipBorderColor;
```
imaNNeoFighT, line 646: I think it's better to have a [BorderSide](https://api.flutter.dev/flutter/painting/BorderSide-class.html) (which contains `color` and `width`) Instead of defining two variables.

lib/src/chart/bar_chart/bar_chart_painter.dart
```diff
@@ -668,6 +672,10 @@ class BarChartPainter extends AxisChartPainter<BarChartData> {
         bottomLeft: radius,
         bottomRight: radius);
     _bgTouchTooltipPaint.color = tooltipData.tooltipBgColor;
+    if (tooltipData.tooltipBorderWidth > 0) {
```
imaNNeoFighT, line 675: Can you please check the opacity of color too here?
Because sometimes people set a transparent color, which we don't need to draw anything.
(Maybe it's better to have an extension function like `borderSide.isVisible()`)

```dart
extension BorderSideExtension on BorderSide {
  bool isVisible() => color.opacity > 0 && width > 0;
}
```

imaNNeoFighT: Please apply it to other places.

lib/src/chart/bar_chart/bar_chart_data.dart
fpplm: good idea

imaNNeoFighT:
Our branching approach has just changed. We had to remove the `dev` branch (Which was your base branch).
Check out our new branching approach [here](https://github.com/imaNNeoFighT/fl_chart/blob/master/CONTRIBUTING.md#you-feel-ready-to-start-okay-now-create-your-branch-from-the-master-branch).

PR #946, opened 2022-03-20
Support tooltip background gradient

martiina:
Supports gradient tooltips

### DIFF
--- a/example/lib/line_chart/samples/line_chart_sample1.dart
+++ b/example/lib/line_chart/samples/line_chart_sample1.dart
@@ -41,7 +41,14 @@ class _LineChart extends StatelessWidget {
   LineTouchData get lineTouchData1 => LineTouchData(
         handleBuiltInTouches: true,
         touchTooltipData: LineTouchTooltipData(
-          tooltipBgColor: Colors.blueGrey.withOpacity(0.8),
+          tooltipBgGradient: const LinearGradient(
+            begin: Alignment.topCenter,
+            end: Alignment.bottomCenter,
+            colors: [
+              Color.fromARGB(255, 255, 230, 0),
+              Color.fromARGB(255, 253, 145, 44),
+            ],
+          ),
         ),
       );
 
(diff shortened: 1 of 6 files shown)


lib/src/chart/bar_chart/bar_chart_painter.dart
```diff
@@ -701,6 +701,12 @@ class BarChartPainter extends AxisChartPainter<BarChartData> {
         bottomRight: radius);
     _bgTouchTooltipPaint.color = tooltipData.tooltipBgColor;
 
+    /// if tooltip gradient is not null then show gradient instead of plain color
+    if (tooltipData.tooltipBgGradient != null) {
```
imaNNeoFighT, line 705: As we reuse this `_bgTouchTooltipPaint`, please write both if and else branches and make `_bgTouchTooltipPaint.shader = null` when we have color, otherwise make `bgTouchTooltipPaint.color = null` or `Colors.transparent`.

example/lib/line_chart/samples/line_chart_sample1.dart
```diff
@@ -41,7 +41,14 @@ class _LineChart extends StatelessWidget {
   LineTouchData get lineTouchData1 => LineTouchData(
         handleBuiltInTouches: true,
         touchTooltipData: LineTouchTooltipData(
-          tooltipBgColor: Colors.blueGrey.withOpacity(0.8),
+          tooltipBgGradient: const LinearGradient(
```
imaNNeoFighT, line 44: This gradient doesn't fit with the line chart sample1 design theme.
Can you please make it look better, or use it on another sample?

image

imaNNeoFighT:
I made an extension function in the **Painter** class called `setColorOrGradient()`
Please check it out and use it like how we used it in the below code:

https://github.com/imaNNeoFighT/fl_chart/blob/846c40ea1698cedbbdc7243f1467ad911aca8a59/lib/src/chart/bar_chart/bar_chart_painter.dart#L331-L336

imaNNeoFighT:
Any progress on this PR?

Issue #969, opened 2022-04-04
Access to RangeAnnotation coordinates


Dnathan33:
** Feature Request**
Feature that provides access to the coordinates of the annotations drawn on the chart (RangeAnnotations). 
The current implementation of RangeAnnotations simply draws a colored container on the chart. This restricts how various applications would like to display annotations. Getting access to the annotation coordinates allows for custom annotations to be drawn.

**Describe the solution you'd like**
A callback function that retrieves the list of annotation coordinates used to draw the annotation container on the chart.
The following PR implements the solution. 

**Additional context**
Here are some use cases where getting access to the RangeAnnotation coordinates can help in rendering custom annotations.
The following was implemented with the changes made on this PR

<img width="398" alt="Screen Shot 2022-04-04 at 9 12 06 AM" src="https://user-images.githubusercontent.com/39175434/161586681-471b6d22-30a3-4aa4-aa79-4904574831b1.png">
<img width="402" alt="Screen Shot 2022-04-04 at 9 12 28 AM" src="https://user-images.githubusercontent.com/39175434/161586743-3b7d2a4a-b39c-4e6c-914a-b73cd51daa80.png">

Dnathan33:
Hey @imaNNeoFighT, I'd appreciate your thoughts on this and PR request.

imaNNeoFighT:
@Dnathan33 I didn't understand the problem.
Where is the custom annotations in your samples?

Dnathan33:
@imaNNeoFighT The custom annotations are drawn under the chart seen in the images attached aboce `|_____|`. The custom annotation is drawn from the Offset coordinates returned.

I have added a sample chart that makes use of the coordinates to draw a custom annotation. Please, look at the attached image and `/line_chart_sample11.dart` for further details.

![Simulator Screen Shot - iPhone 13 - 2022-05-12 at 10 48 40](https://user-images.githubusercontent.com/39175434/168137343-0b11b36a-a3ec-454d-8d84-b080695ba6ab.png)

Issue #801, opened 2021-10-26
How to translate localPosition of TouchEvent to coordinates of the LineChart? (projection)


hmumixaM:
I am working on the LineChart, and I wish to allow users to drag points freely in the chart so that they can perform analysis on it. Analysis includes drawing a line based on 2 points of selection and calculate the length of the line. 

The `LineTouchResponse` of `touchCallback` only returns coordinates of curves with given x-coordinate, but I wish the point can be dragged freely even outside the curve. 

My current approach is to get the coordinate that user pointing at, and `setState` to draw a new point there. I am able to get `localPosition` (the pixel coordinate of the phone) from `FlTouchEvent.localPosition`, but I cannot draw a new point with the pixel coordinate. So is there a method to translate the `localPosition` to chart coordinates?

icefee:
A flexible solution, It's works for me
`Offset pixelToCoordinate(Offset offset, BoxConstraints constraints) {
    return Offset(
        minX + (maxX - minX) * (offset.dx - reservedSizeX) / (constraints.maxWidth - reservedSizeX),
        minY +
            (maxY - minY) *
                (constraints.maxHeight - offset.dy - reservedSizeY) /
                (constraints.maxHeight - reservedSizeY));
  }`

icefee:
A flexible solution, It's works for me

BoxConstraints  provided by a LayoutBuilder wrapped outside the LineChart

```dart
Offset pixelToCoordinate(Offset offset, BoxConstraints constraints) {
    return Offset(
        minX + (maxX - minX) * (offset.dx - reservedSizeY) / (constraints.maxWidth - reservedSizeY),
        minY +
            (maxY - minY) *
                (constraints.maxHeight - offset.dy - reservedSizeX) /
                (constraints.maxHeight - reservedSizeX));
}
```

icefee:
A flexible solution, It's works for me

BoxConstraints  provided by a LayoutBuilder wrapped outside the LineChart

```dart
Offset pixelToCoordinate(Offset offset, BoxConstraints constraints) {
    return Offset(
        minX + (maxX - minX) * (offset.dx - reservedSizeY) / (constraints.maxWidth - reservedSizeY),
        minY +
            (maxY - minY) *
                (constraints.maxHeight - offset.dy - reservedSizeX) /
                (constraints.maxHeight - reservedSizeX));
}
```

imaNNeo:
[...]

PR #970, opened 2022-04-04
AnnotationRange coordinates (Issue: #969)

Dnathan33:
Addresses and resolves issue noted in issue #969


Dnathan33:
cc @imaNNeoFighT

imaNNeoFighT:
Please answer [this comment](https://github.com/imaNNeoFighT/fl_chart/issues/969#issuecomment-1094118601) first.

Dnathan33:
@imaNNeoFighT Addressed your comment

imaNNeoFighT:
What if we handle it in a built-in way?

imaNNeoFighT:
Implementing a generic way to convert the chart's coordinate system to Flutter's coordinate is better.
For example, you have a `_chartController` reference, then you call `_chartController.converToPixels(x: 10, y: 3)` and it returns something equalivent to the Flutter coordinate system.

imaNNeoFighT:
Because something related asked here: #801

Dnathan33:
> Implementing a generic way to convert the chart's coordinate system to Flutter's coordinate is better.
> For example, you have a `_chartController` reference, then you call `_chartController.converToPixels(x: 10, y: 3)` and it returns something equalivent to the Flutter coordinate system.

Screen Shot 2022-10-26 at 11 26 26 AM

@imaNNeoFighT  If you can see from the attached image above (last chart) this PR only adds logic to get the coordinates of the starting and ending point of annotations. It does not generate coordinates based on user interaction and there is no need to convert the chart's coordinate system to Flutter's coordinate.

Dnathan33:
@imaNNeoFighT Following up on this.

imaNNeo:
So we have restructured the underlying drawing logic. Now titles are widgets and I think we can implement this feature using widgets and allow people to put anything they want. Please check the axis_chart_scaffold_widget.dart

https://github.com/imaNNeo/fl_chart/blob/e2ed862bbfaa0931a6fca1b0141d3feb240c37c1/lib/src/chart/base/axis_chart/axis_chart_scaffold_widget.dart#L11-L17

PR #992, opened 2022-05-06
Fix a NAN crash for HTML web renderer

shalehaha:
cc @imaNNeoFighT

We are seeing a NAN crash when used it with HTML web renderer. I have tracked it down and found the line of code that is causing the problem.
I'm not quite familiar with the code base, so I just made a quick fix for it which is easiest for me. There might be a more elegant way to fix it, in that case, you can just close my PR.

### DIFF
--- a/lib/src/chart/base/axis_chart/side_titles/side_titles_widget.dart
+++ b/lib/src/chart/base/axis_chart/side_titles/side_titles_widget.dart
@@ -110,11 +110,13 @@ class SideTitlesWidget extends StatelessWidget {
     double axisMin,
     double axisMax,
   ) {
+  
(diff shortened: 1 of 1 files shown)


imaNNeoFighT:
Hi. Can you please send me a stack trace or something like that?

imaNNeoFighT:
Is there a way to reproduce it?

svki0001:
I have the same issue for mobile browsers, when using SideTitles in a LineChart.

imaNNeoFighT:
This issue has been fixed in [0.45.0](https://github.com/imaNNeoFighT/fl_chart/releases/tag/0.45.0).
For more information, check here:
https://github.com/imaNNeoFighT/fl_chart/blob/42e9863589072c33c0ae1bd9418da698c98eb075/lib/src/utils/utils.dart#L138-L140

PR #1002, opened 2022-05-12
Add 'chartRendererKey' property to LineChart, BarChart, and ScatterChart

### LINKED ISSUE #987: Key/Size of chart area only
Hi I am wondering if there is someway to add a key to the chart area only, without the FLTitles. I am wondering this because I want´to access the Size of the chart are only ( I am using the size in a gesture detector so I can draw line on the chart) . 

If I use the FLTitles in the chart the drawing of the chart gets offset by the width and height of the FLTitles and the subsequent calculation gets offset  as well.  This is sub-optimal as i have to kind of eyeball it and the touched area is not updating correctly.

My current workaround would be to put some 'home made' titles in a row and column around the LineChart and GestureDetector stack.

imaNNeoFighT:
Fixes #987

Check [this gist](https://gist.github.com/imaNNeoFighT/678d5bdfb31421f133c404c1d7be60b2).

https://user-images.githubusercontent.com/7009300/168181809-0164b953-9a5e-4208-989c-4a957ff161f4.mov


hpiu:
nice :)

PR #1057, opened 2022-06-11
Add the ability to rotate the RadarChart titles

FlorianArnould:
Just one question :
Do we need to update the Markdown radar chart doc with more detailed information about the need behavior of the `RadarChartData.getTitle` field
Maybe the migration doc too


imaNNeoFighT:
Everything is good and Approved.

But before merging please do the below requirements.

1. Please update the `radar_chart.md` doc to add `RadarChartTitle` section.

2. Please write about your *BREAKING* change in the `CHANGELOG.md` fil. Also, it would be great if you write a migration guide (something like what we did for `0.50.0` version [here](https://github.com/imaNNeoFighT/fl_chart/blob/master/repo_files/documentations/migration_guides/MIGRATION_00_50_00.md). You can name it `MIGRATION_NEW_VERSION.md` for now. I will rename it whenever I created our next version.

Thanks in advance!

FlorianArnould:
Ok, I added some doc as you asked I'm not sure if it's enough or too much.  
Don't hesitate if there is something else to change, add or delete :wink:

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 * **IMPROVEMENT** (by @imaNNeoFighT): Use Container to draw axis-based charts border.
 * **FEATURE** (by @FlorianArnould) Add the ability to select the RadarChart shape (circle or polygon), #1047.
 * **BUGFIX** (by @imaNNeoFighT): Fix LineChart titles problem with single FlSpot, #1053.
+* **BREAKING** (by @FlorianArnould) Add the ability to rotate the RadarChar titles, #883. This breaks [RadarChartData.getTitle](https://github.com/imaNNeoFighT/fl_chart/blob/master/repo_files/documentations/radar_chart.md#RadarChartData) which have a new parameter `angle` and now returns a [RadarChartTitle](https://github.com/imaNNeoFighT/fl_chart/blob/master/repo_files/documentations/radar_chart.md#RadarChartTitle) instead of a simple `string`. (Read our [Migration Guide](https://github.com/imaNNeoFighT/fl_chart/blob/master/repo_files/documentations/migration_guides/MIGRATION_NEW_VERSION.md) to learn more about it)
```
imaNNeoFighT, line 6: It would be great if we have two separate lines. One for our new **FEATURE**. And another line below that for **BREAKING**. Instead of combining both in a single line.

imaNNeoFighT:
I think it's better to move your images beside the migration file. Because they are fully related to the migration file (in images/*_chart/ we put images that have a corresponding implementation in our example app).
What do you think?

CHANGELOG.md
FlorianArnould: I split it into 2 different lines

FlorianArnould:
> I think it's better to move your images beside the migration file. Because they are fully related to the migration file (in images/*_chart/ we put images that have a corresponding implementation in our example app). What do you think?

Yes seems better to me to avoid flooding the example image folder.

What have you in mind ? something like `repo_files/documentations/migration_guides/images` ?

imaNNeoFighT:
It was a straightforward PR. 
It's my pleasure to see more from you in this package :)

PR #1058, opened 2022-06-13
update bottomTitles method

xzghx:
update bottomTitles method to create bottom titles  without  switch cases.


imaNNeoFighT:
Thank you!

PR #1068, opened 2022-06-20
Feature -  Horizontal scroll implementation line chart example

martinkong0806:
Hello @imaNNeoFighT 

I had made an example to allow users to scroll the graph content, and I think it might be useful to some people, I am not sure if this feature is need to implemented inside the lib, but at least this is how I would be doing it.

Thanks

### DIFF
--- a/example/lib/line_chart/line_chart_page.dart
+++ b/example/lib/line_chart/line_chart_page.dart
@@ -50,7 +50,6 @@ class LineChartPage extends StatelessWidget {
             child: LineChartSample2(),
           ),
           SizedBox(height: 22),
-
           Padding(
             padding: EdgeInsets.only(left: 28.0, right: 28),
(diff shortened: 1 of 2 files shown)


imaNNeoFighT:
Wow! That's amazing. I like your approach to handling this issue. As you might know, we have an ancient issue #71 from two years ago to handle the horizontal scroll.

I think it's better to handle it internally. You can find the AxisChartScaffoldWidget to get started.
Now I checked the Flutter source code. I think we can use the `Scrollable` widget (parent of ListView, GridView, SingleChildScrollView, ....). In this way, we can use the flutter's approach to handle the scroll (interactions, physics, etc).

Let me know what do you think?

Milllli:
I got what it take but I’m still confusing on how to begin Sony wanna make a mistake

Issue #1078, opened 2022-07-01
RadarChart tick and graph point scale do not match


ateich:
**Describe the bug**
RadarChart tick and graph point scale do not match

If a point has a value of 15 and a tick line has a value of 15, I would expect the point to fall on the line. This is not always the case

The values in tick titles do not always match with the values of the points on the graph because
- points are drawn assuming the graph center = 0
- tick title values assume that the graph center = (minValue - ((maxValue - minValue) / tickCount))

**To Reproduce**
```
import 'package:fl_chart/fl_chart.dart';
import 'package:flutter/material.dart';

void main() {
  runApp(const MyApp());
}

class MyApp extends StatelessWidget {
  const MyApp({Key? key}) : super(key: key);

  // This widget is the root of your application.
  @override
  Widget build(BuildContext context) {
    return const MaterialApp(
      home: Scaffold(
        body: Center(
          child: UsageRadarChart(),
        ),
      ),
    );
  }
}

class UsageRadarChart extends StatelessWidget {
  const UsageRadarChart({Key? key}) : super(key: key);

  @override
  Widget build(BuildContext context) {
    return SizedBox(
      width: 300,
      height: 300,
      child:
        RadarChart(
          RadarChartData(
            tickCount: 2,
            dataSets: [
              RadarDataSet(
                dataEntries: [
                  const RadarEntry(value: 15),
                  const RadarEntry(value: 20),
                  const RadarEntry(value: 20),
                ]
              )
            ],
          ),
        ),
    );
  }
}
```

**Screenshots**
Expected | Actual
:-------------------------:|:-------------------------:
Points: [15, 20, 20] | Points: [15, 20, 20]
<img width="245" alt="Screen Shot 2022-07-01 at 7 22 54 PM" src="https://user-images.githubusercontent.com/4775201/176977246-79eb4ad3-0fc4-499b-8684-61e6cca2c847.png"> | <img width="245" alt="Screen Shot 2022-07-01 at 7 22 30 PM" src="https://user-images.githubusercontent.com/4775201/176977255-35a1da33-f8ee-45ad-b46b-522beaf01d30.png">

**Versions**
 - which version of the Flutter are you using?
```
Flutter 3.0.0 • channel stable • https://github.com/flutter/flutter.git
Framework • revision ee4e09cce0 (8 weeks ago) • 2022-05-09 16:45:18 -0700
Engine • revision d1b9a6938a
[...]

PR #1084, opened 2022-07-06
fix: RadarChart tick and graph point scale do not match

ateich:
Fixes an issue where RadarChart tick and graph point scale do not match.

If a point has a value of 15 and a tick line has a value of 15, I would expect the point to fall on the line. This was not always the case.

The values in tick titles did not always match with the values of the points on the graph because
- points were drawn assuming the graph center = 0
- tick title values assumed that the graph center = (minValue - ((maxValue - minValue) / tickCount))

This commit resolves this issue by changing the scaling factor applied to points when they are drawn so that they match the scale of the tick lines.

New unit tests have been added and existing tests that used the incorrect scaling have been updated.

Resolves #1078

PR #1091, opened 2022-07-08
feat(radar-chart): Add minimum/maximum value overrides

### LINKED ISSUE #1090: Allow overriding of Radar Chart min/max values for tick calculation
In cases where the possible minimum/maximum values are known in advance, it should be possible to use these to determine the starting/ending ticks of Radar Charts. This would allow stable and comparable charts across disparate data sets.

Currently, the tick spacing and the supporting titles are calculated by taking the minimum and maximum data entry values, which is fine for the general case, but makes it hard to compare datasets against absolute margins, as the ticks change position according to the entries.

zmeggyesi:
New values are not required, will fall back to previous functionality if omitted.

closes #1090, #1078


zmeggyesi:
@imaNNeoFighT Pubspec intentionally left untouched, in case you want to step versions according to your own schedule.

lib/src/chart/radar_chart/radar_chart_painter.dart
```diff
@@ -96,8 +99,8 @@ class RadarChartPainter extends BaseChartPainter<RadarChartData> {
       canvasWrapper.drawPath(path, _borderPaint);
     }
 
-    final dataSetMaxValue = data.maxEntry.value;
-    final dataSetMinValue = data.minEntry.value;
+    final dataSetMaxValue = data.maxValue ?? data.maxEntry.value;
+    final dataSetMinValue = data.minValue ?? data.minEntry.value;
```
ateich, line 103: I think this will set the first tick value to data.minValue instead of the center of the chart to data.minValue

For example, if minValue = 0, maxValue = 30, tickCount = 2 the center of the chart would be -15 (unlabeled as only the tick lines are labeled), the first tick will be 0 (data.minValue), the second tick would be 15, and the outermost line would be 30 (unlabeled, data.maxValue)

Is this the expected behavior?

zmeggyesi: Huh. No, not really.

But then doesn't this mean the current implementation would also mislabel the center of the chart?

ateich: I don't think the center of the chart is mislabeled in the current implementation, but it can be misleading because of how it labels things

The best guess I can come up with for the existing functionality is that the center was deliberately made to be less than the min value to give the data points some visual distance from the center. I'm not sure if we should keep it this way for the default, but I'm hesitant to change it as it would change how existing graphs are displayed. We could potentially make the change in a new major version release if @imaNNeoFighT thinks it is the right path to take

Adding max and min values for the chart is great since it would not be a breaking change

As for the outermost line being labeled, this is something I would like to see be added in the future. I think it isn't currently labeled as it might overlap with the chart titles

zmeggyesi: So I guess my implementation doesn't actually break anything after all? Given that the centerpoint is, from your comment, _supposed_ to be smaller than `minValue`.  
Interestingly, I just tested this by adding a zero datapoint and deliberately setting my `data.minValue` to `0`, and it looks like the center is _clamped_ to zero:
image

Same thing happens if I _forgo_ setting `data.minValue`, the chart looks exactly the same.

What's your opinion, did I introduce a bug with this PR?

ateich: In the current implementation, [points and tick lines do not always line up as they should](https://github.com/imaNNeoFighT/fl_chart/issues/1078) so I'm not sure if the chart you are showing is completely accurate. In particular, the broken current implementation assumes that the chart center is always 0 when drawing points, but assumes that the center is (minValue - ((maxValue - minValue) / tickCount)) when drawing tick lines. I think the center point in your chart would actually be drawn at the 0.0 tick line in the [fixed version](https://github.com/imaNNeoFighT/fl_chart/pull/1084), which would mean that the center of the chart would actually be at -6.0

Can you send me the code you used to create that chart so I can check it against my branch that has the fix for the existing mismatched points and tick lines issue?

zmeggyesi: Oh, the data points not lining up on the ticks precisely is not an issue here, I just wanted to illustrate that a data point of zero seems to be clamped to the center.

I did have another issue with scaling, which I discovered on a chart where the data _should_ have lined up, but my next commit fixes that - I neglected to propagate the `data.minValue` to the data point scaling function.

lib/src/chart/radar_chart/radar_chart_data.dart
```diff
@@ -172,6 +180,8 @@ class RadarChartData extends BaseChartData with EquatableMixin {
             tickBorderData ?? const BorderSide(color: Colors.black, width: 2),
         gridBorderData =
             gridBorderData ?? const BorderSide(color: Colors.black, width: 2),
+        maxValue = maxValue,
```
imaNNeoFighT, line 183: It's better to keep the maxValue non-nullable (It can be nullable in the constructor, and we can fill the non-nullable maxValue field using this logic: `(data.maxValue ?? data.maxEntry.value);`

imaNNeoFighT: Check the below logic that we have implemented in the LineChartData:

https://github.com/imaNNeoFighT/fl_chart/blob/becd7bda3c9ff6fcc280c78763b438057bc2fa74/lib/src/chart/line_chart/line_chart_data.dart#L92-L97

imaNNeoFighT:
BTW You need to write some unit-tests to pass the codecov pipeline

lib/src/chart/radar_chart/radar_chart_data.dart
zmeggyesi: Thanks for the feedback. The new commits should implement these suggestions.

lib/src/chart/radar_chart/radar_chart_data.dart
```diff
@@ -351,27 +374,30 @@ class RadarEntry with EquatableMixin {
...
   /// Lerps a [RadarDataSet] based on [t] value, check [Tween.lerp].
   static RadarEntry lerp(RadarEntry a, RadarEntry b, double t) {
     return RadarEntry(value: lerpDouble(a.value, b.value, t)!);
   }
+}
 
-  /// Used for equality check, see [EquatableMixin].
-  @override
-  List<Object?> get props => [value];
+enum RadarShape {
```
imaNNeoFighT, line 390: There is a lot of code displacement. Please revert them all.

lib/src/chart/radar_chart/radar_chart_data.dart
```diff
@@ -240,24 +242,45 @@ class RadarChartData extends BaseChartData with EquatableMixin {
     }
   }
 
-  /// Used for equality check, see [EquatableMixin].
+  static _RadarChartMinMaxValues calculateMaxValue(List<RadarDataSet>? datasets) {
```
imaNNeoFighT, line 245: This should be `calculateMinMaxValues`

lib/src/chart/radar_chart/radar_chart_data.dart
zmeggyesi: :maximumfacepalm: I must have formatted the code by reflex. I'm sorry, I'll revert it in the next commit...

imaNNeoFighT: Thank you!

lib/src/chart/radar_chart/radar_chart_painter.dart
```diff
@@ -13,6 +13,9 @@ class RadarChartPainter extends BaseChartPainter<RadarChartData> {
   late Paint _graphPaint, _graphBorderPaint, _graphPointPaint;
   late TextPainter _ticksTextPaint, _titleTextPaint;
 
+  final double? maxValue;
```
imaNNeoFighT, line 16: Now we can remove these local variables.

imaNNeoFighT:
Something is wrong with your code style.
Please run `make format` and `make analyze` to see what's wrong.

zmeggyesi:
Sorry for the delay. I've cleaned up those orphaned variables and acted on the Analyzer warnings. I've also pulled the latest `master`, just to be on the safe side of things.

lib/src/chart/radar_chart/radar_chart_renderer.dart
```diff
@@ -37,7 +37,9 @@ class RenderRadarChart extends RenderBaseChart<RadarTouchResponse> {
       : _data = data,
         _targetData = targetData,
         _textScale = textScale,
-        super(targetData.radarTouchData, context);
+        super(targetData.radarTouchData, context) {
+    painter = RadarChartPainter();
```
imaNNeoFighT, line 41: What is this painter?

imaNNeoFighT:
Any update?

zmeggyesi:
Sorry, I haven't had the bandwidth to fix the CC issue yet. I'll see what I
can do in the near future.

eldrigo:
Is this feature implemented?

imaNNeo:
We had a similar fix here: #1293

PR #1103, opened 2022-07-22
chore: BaseChart and AxisChart is unused

lsaudon:
AxisChart don't have reference.


imaNNeoFighT:
Good catch!
Thank you.

PR #975, opened 2022-04-15
Add ability to use ExtraLinesData in the BarChartData

NikitaPipita:
Resolves issue #476

- Moved ExtraLinesData class and related classes from line_chart_data.dart to axis_chart_data.dart;
- Moved drawExtraLines function from LineChartPainter class to AxisChartPainter class;
- Added extraLinesData variable to AxisChartData and BarChartData constructors.

PR #1109, opened 2022-08-02
Resolved #975 pull request conflicts.

bigFaceNinja:
Thank you for always providing great packages.
I really wanted to use extra lines for bar charts at work, so I resolved the following pull request conflicts.

> Add ability to use ExtraLinesData in the BarChartData #975


imaNNeoFighT:
Hi there. Why you didn't update the main PR and create a new one? 
We need the history of our conversation in the future. That's why it's better to update the main branch. Can you do that?

imaNNeoFighT:
BTW you need to write some unit-tests.

PR #1166, opened 2022-10-12
chore: Add dependabot for Github Actions and Pub

lsaudon:


imaNNeoFighT:
Can you explain what it does?

lsaudon:
Yes, it's a tool for check new version of dependencies.
When a new version is available, Dependabot create a PR with the new version of the dependence.

imaNNeoFighT:
Such a great feature!
Thank you.

PR #1184, opened 2022-11-03
default to 2d eucledian distance

vfolz:
Using 2d euclidean distance allows for touch events to be more precise based on (x,y) coordinates as opposed to the default, which is to include all points along the X-axis. Using this 2d calculation eliminates the issue of tooltips growing beyond the screen and has a minor impact on single line view.

### DIFF
--- a/lib/src/chart/line_chart/line_chart_data.dart
+++ b/lib/src/chart/line_chart/line_chart_data.dart
@@ -1,6 +1,6 @@
 // coverage:ignore-file
 import 'dart:ui';
-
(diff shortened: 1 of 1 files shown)


imaNNeoFighT:
This logic is intentional. A lot of people are working with this default logic at the moment and I didn't see any huge complaints about it.
You can create an issue about it and if it gets a lot of thumbs-up, then we can think about changing the default behavior.
BTW everyone can modify the logic at the moment because it is customizable.
So I will close this PR.

PR #1263, opened 2023-02-14
docs(FlGridData): correct default value of drawVerticalLine

CCXXXI:
It is true since 6328ddc95e472aa4b627f00bf46c2d6a11e72e6e.


imaNNeo:
Thank you!

PR #1266, opened 2023-02-18
FIX: 1225 horizontal lines not rendering at max and min

### LINKED ISSUE #1255: top and bottom `horizontalLines` in `ExtraLinesData` not rendering since `^0.60.0`
Hello,
when I upgraded to `^0.60.0` from `^0.55.2` top and bottom lines are not visible. 

```dart
ExtraLinesData(
        extraLinesOnTop: false,
        horizontalLines: Mood.values
            .mapIndexed(
              (i, _) => HorizontalLine(
                y: i.toDouble(),
                color: NepanikarColors.primary.withOpacity(0.2),
                strokeWidth: 1,
                dashArray: [5],
              ),
            )
            .toList(),
      )
```

Expected / before:

Current:

Full code of chart since the code is open sourced: https://github.com/cesko-digital/nepanikar/blob/main/lib/widgets/mood/mood_chart.dart

JoshMart:
This fixes #1255 where extra horizontal lines would not be painted when they were to be drawn horizontally along the min or max of the chart.


lib/src/chart/base/axis_chart/axis_chart_painter.dart
```diff
@@ -214,11 +214,9 @@ abstract class AxisChartPainter<D extends AxisChartData>
       final from = Offset(0, getPixelY(line.y, viewSize, holder));
       final to = Offset(viewSize.width, getPixelY(line.y, viewSize, holder));
 
-      final isLineBeingDrawnOutsideChart = from.dy <= 0 || to.dy <= 0;
-      final isLineBeingDrawnOnXAxis =
-          from.dy == viewSize.height || to.dy == viewSize.height;
+      final isLineBeingDrawnOutsideChart = from.dy < 0 || to.dy < 0;
 
-      if (!(isLineBeingDrawnOutsideChart || isLineBeingDrawnOnXAxis)) {
+      if (!isLineBeingDrawnOutsideChart) {
```
imaNNeo, line 219: What if we remove the whole if statement? Because I don't see any if statement in the `drawVerticalLines` function.
BTW if you want to keep the restriction for drawing lines inside the chart, I think you need to check both bottom and top bounds:

```dart
final isLineBeingDrawnOutsideChart = from.dy < 0 || to.dy < 0 || from.dy > viewSize.height || to.dy > viewSize.height;
...
```

Also please update the `drawVerticalLines` function to restrict the drawing between `0` and `viewSize.width`

imaNNeo:
We have this logic in BarChart too. Please check and fix it if there is any bug in the BarChart.
Thanks!

lib/src/chart/base/axis_chart/axis_chart_painter.dart
JoshMart: I'd be okay removing the whole if statement. It was my attempt at preventing lines being drawn off the chart. But...maybe we should just leave that up to the developers to ensure they aren't drawing any lines off the chart instead? If you agree I'll remove this check and won't make any updates to `drawVerticalLines`.

imaNNeo: In my opinion, it is better to restrict it. Because there are some issues that developers complained about why the line draws out of the chart.
Can we check it with the chart's `minX`, `maxX`, `minY`, and `maxY`?

JoshMart: Sounds good. I'll look into restricting drawing of extra lines inside of the bounds. I'm about to head on some PTO so I will not be around for about a week.

imaNNeo: Great!
Take your time.

JoshMart: > Great! Take your time.

Just pushed up some changes. Extra lines are now being drawn on x and y min/max values and are ignored if they are less than min or greater than max.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 * **BUGFIX** (by @imaNNeo) Show `0` instead of `-0` in some edge-cases in the default titles 
 * **FEATURE** (by @tamasapps): Add `tooltipHorizontalAlignment` and `tooltipHorizontalOffset` property in [LineTouchTooltipData], [BarTouchTooltipData], [ScatterTouchTooltipData].
 * **FEATURE** (by @dhiyaaulauliyaa) Add ability to force SideTitle to be placed inside its corresponding axis bounding box, #603.
+* **BUGFIX** (by @JoshMart) Fix horizontal lines not painting when at chart min or max, #1255.
```
imaNNeo, line 7: 0.61.0 was the previous release.
Please read the [contribution guideline](https://github.com/imaNNeo/fl_chart/blob/master/CONTRIBUTING.md) carefully.

You need to create a new release on the top called `newRelease`. Like the below code:

```
## newRelease
* **BUGFIX** (by @JoshMart) Fix horizontal lines not painting when at chart min or max, #1255.
```

imaNNeo:
BTW everything looks good to me.
Please fix the codecov issue, then we are ready to merge it.

CHANGELOG.md
JoshMart: Good catch, thanks. Got this updated.

JoshMart:
> BTW everything looks good to me. Please fix the codecov issue, then we are ready to merge it.

`codecov/patch` is reporting uncovered lines on code that existed before this PR. The lines in question were simply indented because they are now wrapped in an if block. They are not new.

imaNNeo:
Okay I have merged it. Thank you!

Issue #1106, opened 2022-07-29
Memory leaking with changing spot list


Clon1998:
**Describe the bug**
When chaning the complete FLSpot list instance/reference e.g. calculating it based on another list whenever the list changes the `LineChart` is leaking a ton of `FlSpots`

Here is a DevTool-measurement within seconds the Spots grows to 100k+ instances ( Example code used to measure can be found below)
<img width="705" alt="Bildschirmfoto 2022-07-29 um 19 27 23" src="https://user-images.githubusercontent.com/10357775/181812733-a93e3061-1586-4a69-8d9a-d6ea2bcf99c8.png">

**To Reproduce**
To debug the behavior I had a look at the [Sample#10](https://github.com/imaNNeoFighT/fl_chart/blob/master/example/lib/line_chart/samples/line_chart_sample10.dart) which does not leak any memory. However, I edited and also managed to reproduce the FLSpot leaking I encountered in my own app.
Instead of using an instance of `final sinePoints=<FlSpot>[]` ([src](https://github.com/imaNNeoFighT/fl_chart/blob/d9d3aa81073b3b3f6029a4fd88388bf02477bce2/example/lib/line_chart/samples/line_chart_sample10.dart#L19)) I am using a `final sinPoints = <double>[];` list which I later map to `FlSpots` using the `_convertToPlotSpots` method.
Also to verify that the issue is caused by the chart I replaced the `LineChart` widget with an `Text('Test:${_convertToPlotSpots(sinPoints).last.x}');` widget that uses also the FlSpot list just to see if the leaking is caused by the code in `_convertToPlotSpots` which is not the case.
Finaly I found a workaround for me by using a single instance of `List<FlSpot>`that does not change/only swapps elements using this code for the `_convertToPlotSpots` method:
```dart
// Workaround that uses again a fixed instance of List<FlSpot> and just replace the elements of it
List<FlSpot> l = [];
List<FlSpot> _convertToPlotSpots(List<double>? doubles) {
  if (doubles == null) return const [];
  List<double> sublist = doubles.sublist(max(0, doubles.length - 300));
  l.clear();
  double i = 0;
  l.addAll(sublist
      .map((e) => FlSpot(i++, e)));
  return l;
}
```

Sample#10 leaking Memory:
```dart
// Widget that leaks memory 
class LineChartSample10 extends StatefulWidget {
  const LineChartSample10({Key? key}) : super(key: key);

  @override
[...]

PR #1283, opened 2023-03-15
add `isEnableCache` option, fix memory lake

wenjiasen:
add `isEnableCache` option, fix memory lake on a dynamic line chart #1106

### DIFF
--- a/lib/src/chart/line_chart/line_chart_data.dart
+++ b/lib/src/chart/line_chart/line_chart_data.dart
@@ -58,6 +58,7 @@ class LineChartData extends AxisChartData with EquatableMixin {
     double? maxY,
     super.baselineY,
     FlClipData? clipData,
+    bool isEnableCache = true,
     super.backgroundColor,
   })  : lineBarsData = lineBarsData ?? const [],
         betweenBarsData = betweenBarsData ?? const [],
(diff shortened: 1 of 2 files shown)


wenjiasen (commented):
When the data of the chart is changing with time, `_cachedResults` in `calculateMaxAxisValues` will not be able to free the data of the old time, which will inevitably lead to a memory leak, and in the absence of a better solution, at least provide a switch to avoid an OOM.

imaNNeo:
It is a trade-off between memory usage and CPU usage.
At this time we decided to sacrifice memory which leads to memory leaks.
If we put a flag, we sacrifice the CPU (a duplicate calculation happens every time). So nothing fixed.
Actually, I am considering deleting the cache system and dealing with the root issue (recalculation issue). There might be some ways to do that.
For example, we can make all of the minX, maxX, minY, and maxY nullable and calculate them in the chart widgets, so we can have a manageable memory and it removes when widget disposes (no more static variables)

What do you think?

wenjiasen:
Thank you for your response. I think the priority of avoiding OOM would be higher than CPU usage, and it might be necessary to provide a way to prevent OOM in dynamic lists before the cache system is fully optimized.

Unexpected crashes may bring unknown risks in scenarios where long-term operation is required, and could cause unnecessary trouble in production environments.

DionaSoft:
I also experience this problem wit flutter_bloc
  @override
    return BlocBuilder<GraphBloc, GraphState>(
      builder: (context, state) {
        return LineChart(
                            LineChartData(
                              lineBarsData: [
                                LineChartBarData(
                                  spots: state.spots,   // List<FlSpot> spots;
                                  ......
                                ),
                              ],
                              .......
                            ),
                            .......
                          ),
      }
}
 It causes heavy memory leaks.
 version 0.61.0

PR #1316, opened 2023-04-17
refactor: Put constructors to const 

lsaudon:
Which I did for the most part.

1. Replace  

```dart
LineChartData({
    List<LineChartBarData>? lineBarsData,
  })  : lineBarsData = lineBarsData ?? const [],
```
to
```dart
LineChartData({
    this.lineBarsData = const [],
  });
```

2. Put builders in const if possible
3. Fix CI

:boom: Rename `drawBehindEverything` to `drawBelowEverything`

### DIFF
--- a/example/lib/presentation/samples/bar/bar_chart_sample1.dart
+++ b/example/lib/presentation/samples/bar/bar_chart_sample1.dart
@@ -226,10 +226,10 @@ class BarChartSample1State extends State<BarChartSample1> {
       ),
       titlesData: FlTitlesData(
         show: true,
-        rightTitles: AxisTitles(
+        rightTitles: const AxisTitles(
           sideTitles: SideTitles(showTitles: false),
         ),
-        topTitles: AxisTitles(
+        topTitles: const AxisTitles(
           sideTitles: SideTitles(showTitles: false),
         ),
         bottomTitles: AxisTitles(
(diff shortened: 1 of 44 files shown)


imaNNeo:
Do you think the new style is better? Why?

Because a long time ago, we were using this style:
```dart
LineChartData({
    this.lineBarsData = const [],
  });
```

Then we decided to change everything to this style:

```dart
LineChartData({
    List<LineChartBarData>? lineBarsData,
  })  : lineBarsData = lineBarsData ?? const [],
```

One of the benefits that I remember is that you can manipulate the input and assign another variable based on your logic in the constructor.

lsaudon:
> One of the benefits that I remember is that you can manipulate the input and assign another variable based on your logic in the constructor.

Yes, it's a benefit for manipulating the input before assign variable.
I have left this way of doing in case we need, but this way of doing things does not put constructors in constant.

The goal is to have constructors in constant if possible, but if there are constraints on the plugin.

imaNNeo:
Great!
So I got convinced, I love const variables.

Please rebase it and remove the draft. I will merge afterward.

Thanks!

lsaudon:
Yes, I was waiting to hear if you agreed with this modification.

lsaudon:
I'm trying to find all the places where we can pass in const or do it in a next pull request ?

imaNNeo:
Let's merge it faster to prevent any conflict for other people,
Then please do the new changes in a new MR.

Thanks a lot.

PR #1325, opened 2023-04-30
class PolygonDotPainter is created that extends FlDotPainter. It ta…

Sardor6628:
…kes in parameters such as sides (number of sides of the polygon), color, stroke width, and radius to customize the polygon shape. The draw method is overridden to implement the polygon shape using the provided parameters. The getSize method is also overridden to return the size of the dot shape based on the radius.

### DIFF
--- a/lib/src/chart/line_chart/line_chart_data.dart
+++ b/lib/src/chart/line_chart/line_chart_data.dart
@@ -1,4 +1,5 @@
 // coverage:ignore-file
+import 'dart:math';
 import 'dart:ui';
 
 import 'package:equatable/equatable.dart';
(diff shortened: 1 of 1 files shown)


Sardor6628:
Screenshot 2023-04-27 at 12 10 36 PM

lib/src/chart/line_chart/line_chart_data.dart
```diff
@@ -982,6 +983,66 @@ class FlDotSquarePainter extends FlDotPainter {
...
+  }
+
+  @override
+  Size getSize(FlSpot spot) {
+    return Size(radius * 2, radius * 2);
+  }
+
+  @override
+  // TODO: implement props
+  List<Object?> get props => throw UnimplementedError();
```
imaNNeo, line 1044: Please implement it, you just need to pass the properties that you have (see other classes)

imaNNeo:
I think you need to write some unit-tests for this feature,
Also please rebase your branch.

Thanks!

PR #1334, opened 2023-05-10
Fixed Link on Example Readme

BrayFlex:
Just a link typo fix (from github, links without the scheme go to `https://github.com/imaNNeo/fl_chart/blob/master/example/app.flchart.dev`), thanks for the library!


imaNNeo:
Thank you very much 😊

PR #1372, opened 2023-06-20
feat: add title to BarChartRodStackItem

### LINKED ISSUE #1371: add title property to BarChartRodStackItem
In the touchCallBack function of BarChart, we can only get the index and value of the currently touching BarChartRodStackItem, but not the name of the value. However, it’s hard to get the name corresponding to the value directly through the index, because the 0 value will not be displayed in the chart. This means that some BarChartRodStackItems have a value of 0, the indexes that originally belongs to them are assigned to other BarChartRodStackItems whose value is not 0. So we can’t directly get the name corresponding to the value/index of BarChartRodStackItem.

Add a title property to the BarChartRodStackItem so that when we touch the BarChartRodStackItem, we can easily get the title that corresponds to its value.

**Describe alternatives you've considered**
Before adding the title property to the BarChartRodStackItem, I had used a calculation function to skip items with a value of 0 and get the correct item. However this is not convenient for dynamic arrays and means that a function needs to be added synchronously for each new scene to maintain it.
截屏2023-06-20 14 06 00

ImagineBoom:
Closes #1371

### DIFF
--- a/lib/src/chart/bar_chart/bar_chart_data.dart
+++ b/lib/src/chart/bar_chart/bar_chart_data.dart
@@ -453,11 +453,11 @@ class BarChartRodStackItem with EquatableMixin {
     this.toY,
     this.color, [
     this.borderSide = Utils.defaultBorderSide,
-    this.title = '',
+    this.title,
   ]);
 
   /// Renders a Stacked Chart section from [title]
-  final String title;
+  final dynamic title;
 
   /// Renders a Stacked Chart section from [fromY]
   final double fromY;
--- a/lib/src/chart/bar_chart/bar_chart_data.dart
+++ b/lib/src/chart/bar_chart/bar_chart_data.dart
@@ -453,7 +453,11 @@ class BarChartRodStackItem with EquatableMixin {
     this.toY,
     this.color, [
     this.borderSide = Utils.defaultBorderSide,
+    this.title = '',
   ]);
+
+  /// Renders a Stacked Chart section from [title]
+  final String title;
 
   /// Renders a Stacked Chart section from [fromY]
   final double fromY;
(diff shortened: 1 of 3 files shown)


imaNNeo:
It would be better if we keep a generic type of data (not only strings). For example, we can keep a `dynamic` type which could be anything

ImagineBoom:
Thank you😀, I have made the changes.

ImagineBoom:
> It would be better if we keep a generic type of data (not only strings). For example, we can keep a `dynamic` type which could be anything

Thank you😀, I have made the changes.

ImagineBoom:
> It would be better if we keep a generic type of data (not only strings). For example, we can keep a `dynamic` type which could be anything

Thank you😀, I have made the changes.

imaNNeo:
`title` is not a good name anymore as it is a dynamic field.
I suggest renaming it to something like `payload` or `data`.
Also, you need to rebase your branch and write some unit tests for it (read our [contributing guidelines](https://github.com/imaNNeo/fl_chart/blob/master/CONTRIBUTING.md) carefully).

lib/src/chart/bar_chart/bar_chart_data.dart
```diff
@@ -453,8 +453,12 @@ class BarChartRodStackItem with EquatableMixin {
     this.toY,
     this.color, [
     this.borderSide = Utils.defaultBorderSide,
+    this.title,
   ]);
 
+  /// Renders a Stacked Chart section from [title]
```
imaNNeo, line 459: Also please update this comment. It seems you copied it from another field

imaNNeo:
Closing, because it has no activity for a while

PR #1387, opened 2023-07-14
Fix BarChart isn't fit to the visibility of its

sownt:
It's cause of `minY` value from [`BarChartHelper`](https://github.com/imaNNeo/fl_chart/blob/master/lib/src/chart/bar_chart/bar_chart_helper.dart) always `0.0` if any `fromY` of `BarChartData` is positive. So min value of Y-axis always starts from `0.0` or a negative value.

I update `minY` value of chart to min of <s>`fromY`</s> any values of `ChartData`. This is result of my change, `Line 1` will be hide and Y-axis will start from 150.

Frame 1

<s>I also update `bar_chart_painter_test` for its.</s>

### DIFF
--- a/lib/src/chart/bar_chart/bar_chart_helper.dart
+++ b/lib/src/chart/bar_chart/bar_chart_helper.dart
@@ -34,7 +34,7 @@ class BarChartHelper {
     }
 
     var maxY = barGroup.barRods[0].toY;
-    var minY = 0.0;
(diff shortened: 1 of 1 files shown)


imaNNeo:
I did some changes to [this (merged) MR](https://github.com/imaNNeo/fl_chart/pull/1422)
Please check the master branch and let me know what do you think about it.

Issue #199, opened 2020-02-11
Gauge chart??


Spiderbezno:
Any options to do a gauge chart?

imaNNeoFighT:
Not yet, but edit the title to more descriptive title, and let it be open,
If it gained lot of thumbs up, we will consider it.
Thanks!

eherediadev:
Adding to the conversation, here we have a couple of designs to consider:

[Gauge meter Vector - Dribbble](https://dribbble.com/shots/6658018-Gauge-meter-Vector-freebie)
[Freebie Smart Home - Dribbble](https://dribbble.com/shots/3081445-Freebie-Smart-Home-App-Sketch)

eherediadev:
@Spiderbezno please change the Title to something like: "Add gauge visualization chart to library".

Regards,

imaNNeoFighT:
Thanks!

imaNNeoFighT:
It is similar to our PieChart, right?

Digoun:
> It is similar to our PieChart, right?

To add on this, a simple design could be similar to the actual PieChart but with a custom angle, by passing optional parameters **startAngle* * and a **endAngle**.
Something similar to the way it done in the [ApexChart](https://apexcharts.com/) library in this example:
https://apexcharts.com/docs/chart-types/radialbar-gauge/#custom-angle

ClecknerT:
Gauges would be super useful addition!

FlorianArnould:
Hi,
I started working on this over the weekend, and I already have a first preview to discuss on (my fork branch [here](https://github.com/FlorianArnould/fl_chart/tree/feature/gauge-chart)). 

For the moment a full example would look like this : 
```dart
GaugeChart(
  GaugeChartData(
    value: 0.7, // double in range [0, 1]
    valueColor: Colors.yellow,
    backgroundColor: Colors.orange,
    strokeWidth: 30,
    startAngle: -90, // degree as you can imagine
    endAngle: 270, // same
    strokeCap: StrokeCap.round, /// can be [StrokeCap.butt] (good looking square), [StrokeCap.round] or [StrokeCap.square] (bad looking square)
  ),
),
```
### strokeCap
<table>
  <tr>
    <th><b>StrokeCap.butt</b></th>
    <th><b>StrokeCap.round</b></th>
    <th><b>StrokeCap.square</b></th>
  </tr>
  <tr>
[...]

PR #1424, opened 2023-08-22
Gauge Chart

FlorianArnould:
The implementation of the gauge chart (discussed here #199)


lib/src/chart/gauge_chart/gauge_chart_data.dart
```diff
@@ -0,0 +1,268 @@
...
+        assert(radius > 0, 'radius should be > 0');
+  final int count;
+  final double radius;
+  final Color color;
+  final GaugeTickPosition position;
+  final double margin;
+  final bool showChangingColorTicks;
+
+  static GaugeTicks? lerp(GaugeTicks? a, GaugeTicks? b, double t) {
+    // TODO(FlorianArnould): if showChangingColorTicks are different
```
imaNNeo, line 134: Is this to do necessary?

lib/src/chart/gauge_chart/gauge_chart_painter.dart
```diff
@@ -0,0 +1,267 @@
...
+  }
+
+  @visibleForTesting
+  Offset center(Size size) => Offset(size.width / 2.0, size.height / 2.0);
+
+  @visibleForTesting
+  double gaugeRadius(Size size) => size.shortestSide / 2;
+}
+
+class GaugePosition {
```
imaNNeo, line 196: Please make this class private (as we need this object only here)

lib/src/chart/gauge_chart/gauge_chart_painter.dart
```diff
@@ -0,0 +1,267 @@
...
+      start = angleSize < 0 ? start + bonusAngle : start - bonusAngle;
+      end = angleSize < 0 ? end - bonusAngle : end + bonusAngle;
+    }
+    return _calculateValidValueRange().contains(vector.distance) &&
+        DegreeAngleRange(start, end)
+            .contains(Utils().degrees(vector.direction));
+  }
+}
+
+class Range {
```
imaNNeo, line 243: Also this one

imaNNeo:
When I put tick count: 3, I see 4 thicks. Is it intentionally?
image

lib/src/chart/gauge_chart/gauge_chart_data.dart
```diff
@@ -0,0 +1,268 @@
...
+  final Color color;
+
+  @override
+  Color getColor(double value) => color;
+
+  @override
+  List<Object?> get props => [color];
+}
+
+class VariableGaugeColor
```
imaNNeo, line 68: This is a confusing approach in this project. Because we didn't use this approach on the other charts.
I would prefer to have a callback to get the progress and return a color (what we do in other charts).
Or a list of colors if you want to handle the gradient.
By the way, I can add these kinds of improvements if you allow me

example/lib/presentation/samples/gauge/gauge_chart_sample1.dart
```diff
@@ -0,0 +1,63 @@
...
+    return Padding(
+      padding: const EdgeInsets.all(28),
+      child: Column(
+        children: [
+          SizedBox(
+            width: 250,
+            height: 250,
+            child: GaugeChart(
+              GaugeChartData(
+                value: _value,
```
imaNNeo, line 27: I think we can have a list of values (each value can have startAngle and endAngle). 
Then we can draw them on top of each other. 
This way, it is more customizable.

Look at these samples:

image

image

image

lib/src/chart/gauge_chart/gauge_chart_data.dart
FlorianArnould: Yes if we want a smooth transition on the color ticks
But would be deleted if we go for another approach like discussed here 
https://github.com/imaNNeo/fl_chart/pull/1424#discussion_r1326595223

FlorianArnould:
> When I put tick count: 3, I see 4 thicks. Is it intentional? image

The 3 grey's are the ticks you asked and the colored one is the one showed because of the showChangingColorTicks on GaugeTicks, this defaults to true

lib/src/chart/gauge_chart/gauge_chart_data.dart
FlorianArnould: The idea is to be able to display some colored ticks at the angle the color changes. 
If you work with a callback, you cannot get this angle precisely and moreover your will need to call it a lot of times with values between 0 and 1 to scan when the color changes.

imaNNeo:
Can you update your branch, please? And reply to the other discussions?

Sun3:
Any updates to adding the Gauge Chart?

Thank you.

matecode:
@FlorianArnould & @imaNNeo Thank you for your work on this topic. How can I help to push the Gauge Chart to the library?

PR #1428, opened 2023-08-26
Feature: Add draggable spots functionality for LineChart

### LINKED ISSUE #1247: Can be possible to add a gesture to drag up and down?
Screenshot 2023-01-31 at 4 56 51 PM
like the circle can be drag to UP and DOWN? By Touching those little circles

or i can be able to this on current package right now?

### LINKED ISSUE #1420: Is it possible to change the shape of the curve by dragging the points of the curve?
example: https://echarts.apache.org/examples/zh/editor.html?c=line-draggable

G33kFreak:
Hello!
First of all, thank you and all contributors for this awesome package, you rule :fire:

This PR closes #1420, closes #1247,  I believe this package definitely needs this type of feature. So here is my approach to achieve it.

## What's done
- Added parameter `isDraggable` for the `LineChartBarData`.
- Added `dragSpotUpdateFinishedCallback` for the `LineTouchData` which returns `UpdatedDragSpotsData`. It allows us to do some logic after dragging has been finished.
- If there is any `LineChartBarData` with an `isDraggable` flag, we cache `lineBarsData` inside `_LineChartState` and update them if we need to, depending on `FlTouchEvent`.
- Added `didUpdateWidget` for `_LineChartState` in order to react to updates in the widget's configuration changes.
- Added a simple sample (preview of it is down below)

## Preview
スクリーンショット 2023-01-25 12 13 31

##
Please check if this approach fits for you and let me know, I'll finish it up and add more docs, also feel free to suggest what's can be done better :)


dateolive:
Great, this is the effect I imagined, this is really great.

G33kFreak:
> Great, this is the effect I imagined, this is really great.

Thanks a lot, I'm still waiting for updates from the maintainer, in meanwhile, you can use my fork and in case of any bugs feel free to report them here :)

techouse:
@G33kFreak holy cow, this is awesome 🚀

mehroze-zaidi:
Great man, I really wanted that feature as this is my requirement.

imaNNeo:
Thank you for this great contribution!
I will add some comments now:

lib/src/chart/line_chart/line_chart.dart
```diff
@@ -121,6 +185,23 @@ class _LineChartState extends AnimatedWidgetBaseState<LineChart> {
...
     });
+
+    // If there is needed event and any lineBar with .isDraggable flag exists,
+    // sets indexes of needed spot and starts dragging process.
+    if (event is FlPanStartEvent || event is FlLongPressStart) {
+      final barIndex = touchResponse.lineBarSpots?.first.barIndex;
+      final spotIndex = touchResponse.lineBarSpots?.first.spotIndex;
+
+      if (spotIndex != null && barIndex != null) {
+        final isDraggable = widget.data.lineBarsData[barIndex].isDraggable;
```
imaNNeo, line 196: I'm a bit confused about the design, we have an `isDraggable` flag in our LineChartBarData class, also we have `DragSpotUpdateFinishedCallback? dragSpotUpdateFinishedCallback;` in our LineTouchData.
 (I know we might have something similar in other places, and I admit that is a bad design)

But for now, what if we keep these two variables beside each other in the LineTouchData class?
What do you think?

imaNNeo: Also, we can add other kinds of callbacks for example `dragSpotStartedCallback`, `dragSpotMovedCallback`, and `dragSpodFinishedCallback`. 

The other (perfect) solution is to touch the [RenderBaseChart](https://github.com/imaNNeo/fl_chart/blob/master/lib/src/chart/base/base_chart/render_base_chart.dart) to add a [DragGestureRecognizer](https://api.flutter.dev/flutter/gestures/DragGestureRecognizer-class.html) to detect the gesture events using this predefined, then we can pass down the gesture events to the touchCallback event as a `FlTouchEvent`. This way, we can use the `drag-drop` feature in all kinds of charts with the same logic.

For example in LineChart, we can have a flag to do the spot drag-drop (that you already implemented) as an under-the-hood magic. Also, users can implement any other kind of interaction with the `drag-drop` feature.

What do you think?

G33kFreak: Thanks for comments! I like the 2nd idea and I'll try to implement it asap :)

G33kFreak: > The other (perfect) solution is to touch the [RenderBaseChart](https://github.com/imaNNeo/fl_chart/blob/master/lib/src/chart/base/base_chart/render_base_chart.dart) to add a [DragGestureRecognizer](https://api.flutter.dev/flutter/gestures/DragGestureRecognizer-class.html) to detect the gesture events using this predefined, then we can pass down the gesture events to the touchCallback event as a `FlTouchEvent`. This way, we can use the `drag-drop` feature in all kinds of charts with the same logic.
> 
> For example in LineChart, we can have a flag to do the spot drag-drop (that you already implemented) as an under-the-hood magic. Also, users can implement any other kind of interaction with the `drag-drop` feature.
> 
> What do you think?

So, after some checks about the additional gesture recognizer, I realized, that it's already done :D

See, the [DragGestureRecognizer](https://api.flutter.dev/flutter/gestures/DragGestureRecognizer-class.html) is an abstract class.
```dart
abstract class DragGestureRecognizer extends OneSequenceGestureRecognizer {
    // ...Its code her
}
```

However, the package already implements [PanGestureRecognizer](https://api.flutter.dev/flutter/gestures/PanGestureRecognizer-class.html) in the [RenderBaseChart](https://github.com/imaNNeo/fl_chart/blob/master/lib/src/chart/base/base_chart/render_base_chart.dart)  which extends [DragGestureRecognizer](https://api.flutter.dev/flutter/gestures/DragGestureRecognizer-class.html) and has everything we need to implement some kind of similar features in other charts as well.

Based on that, we already have the needed gesture recognizer and this PR just creates the wanted `under-the-hood magic.` for the `LineChart`.


> I'm a bit confused about the design, we have an `isDraggable` flag in our LineChartBarData class, also we have `DragSpotUpdateFinishedCallback? dragSpotUpdateFinishedCallback;` in our LineTouchData. (I know we might have something similar in other places, and I admit that is a bad design)
> 
> But for now, what if we keep these two variables beside each other in the LineTouchData class? What do you think?

To be honest, I don't really think what would be better here. My intention was to allow users to implement a draggable feature for particular `LineChartBarData`. For example, we can have a few bars, but only one of them is going to be draggable. It's similar to `isCurved` flag to me.

```dart
lineBarsData: [
    LineChartBarData(
        spots: spots1,
        isCurved: true,
        isDraggable: true,
    ),
    LineChartBarData(
        spots: spots2,
        isCurved: true,
        isDraggable: false,
        color: Colors.red,
    ),
],
```

So that's why I put this flag there.

> Also, we can add other kinds of callbacks for example `dragSpotStartedCallback`, `dragSpotMovedCallback`, and `dragSpodFinishedCallback`.
> 

Definitely! Added it in the commit f42ba9986be9f2833c1ed9b93cea4ef43cb63cba.
Anyway, we have to pay users' attention to be careful with using `setState` in the `dragSpotUpdateCallback`. Would be good to have some notes in the docs :)

imaNNeo:
Can you please update your branch? Then I can take a look again.
Thanks!

G33kFreak:
> Can you please update your branch? Then I can take a look again. Thanks!

I'll do it as soon as I have some free time :)

G33kFreak:
Well, it's been a while since I created this PR (shame on me), however, rebased it for you so now you can take a look, I'll try my best to finish it this time :)

imaNNeo:
Hi, it seems there is a conflict on CHANGELOG.md. You don't need to update the CHANGELOG.
Also, there are 152 commits on your branch. Are you sure you want to keep them all?

imaNNeo:
Closing, because it had no activity for a while

PR #1445, opened 2023-09-22
Added initial animation for LineChart / BarChart / RadarChart

Dartek12:
PR adds some animations at the first appearance of the widgets: LineChart / BarChart / RadarChart. It is enabled by default but can be disabled by providing simple configuration object.
For line/bar charts it works by replacing displayed data with datasets where y axis values are zeroed (most of the time) so that they are animated from the bottom of the graph by evaluating the tween up to the target/initial data.
For radar chart a _scaleFactor_ value was introduced for the renderer so that it can multiply the target distance from the center of the radar to the target point. It is used in the initial animation by moving from _scaleFactor_ 0 to 1.

Issue: https://github.com/imaNNeo/fl_chart/issues/406


jebstern:
If data changes, will e.g Bar values jump to zero, then animate to new values?

Dartek12:
@jebstern That should not happen based on this PR, it would animate from current values to the next ones

lib/src/chart/base/base_chart/initial_animation_mixin.dart
```diff
@@ -0,0 +1,55 @@
...
+      return data;
+    }
+    _targetValue = data;
+
+    return getAppearanceAnimationData(data);
+  }
+
+  T getAppearanceAnimationData(T data);
+
+  double getAppearanceValue(double minY, double maxY) {
```
imaNNeo, line 47: We have PieChart and RadarChart which are not axis-based charts, so we need to implement an approach that supports all kinds of charts. (In the future, we will add more chart types which are not axis-based. Such as [GaugeChart](https://github.com/imaNNeo/fl_chart/pull/1424))

imaNNeo:
Hi, Thanks for contributing!
I think this approach is not customizable, for example, suppose I want to animate the x-axis of my line chart. Or suppose I want to animate the values of my PieChart sections, or ...
This feature (initial animation) is one of the most wanted features by the users and if we merge this approach, it would be hard to develop a new approach and force users to migrate to the new approach (Otherwise we need to handle backward compatibility)

That's why I write the points that are on my head here, then you can try to implement them:

* Animation starts when the widget is created (for the first time) and can be disabled by the user by passing something like an [AnimationController](https://api.flutter.dev/flutter/animation/AnimationController-class.html). In this way, the user needs to show it manually when the widget is created or run it whenever they want.
* It should be highly customizable (as we develop a highly customizable chart library). Users can override a method|callback to customize the animation. For example, they can animate the x-axis of the line chart, or animate each line separately in a sequence. Or I might want to animate the radius of my PieChart. These can be some pre-defined animations which user can choose from. Also, they can create a new one (or maybe they can contribute the created one to add as a pre-defined animation)
* The user should be able to handle the animation and it should be different from the swapAnimation which runs whenever data changes (color, size,...)

Issue #1307, opened 2023-04-11
Tapping outside of pie chart intercepts the gesture


jtkeyva:
** Don't make a duplicate issue.
You can search in issues to make sure there isn't any already opened issue with your concern.

**Describe the bug**
I have 2 pie charts stacked on top of eachother. I cannot intercept the gesture on the bottom pie chart. I set up a print statement to get the index for each segment on the top one and i notice it prints -1. So that tells me the transparent area behind the piechart is intercepting the getstures.

**To Reproduce**
Put a pie chart on top of a box (or any other widget) of the same width as the pie chart that has a gesture recognizer. Try and get the box underneath to respond to touches just outside of the circular pie chart. The box underneath does not respond.

**Screenshots**
If applicable, add screenshots, or videoshots to help explain your problem.

**Versions**
Latest
Latest

aabirsark:
I think you can solve this problem by wrapping IgnorePointer() on the top pie chart !

jtkeyva:
@aabirsark thanks, but wouldn't that disable any interaction on the pie chart itself? i'm looking to be able to tap each slice of the pie as well as the widget that surrounds it

aabirsark:
> @jtkeyva what I would do in that case is too use a clipper and clip the piechart till you want the interaction

jtkeyva:
ah so clip will allow taps to pass through the circular edges?

samir7osny:
check this PR #1504

PR #1504, opened 2023-11-26
Add external touch handler for Pie Chart

samir7osny:
Regarding the case of stacked pie charts on each other, the touch handler will work on one of them only, check this issue #1307 as well.

To try this case and understand the issue, you can go to this commit ceb4e18 and check sample 4.
image

This PR fixes this by adding the ability to calculate the touch response based on external parameter for the touched position, now you can use MouseRegion widget for example and add the stacked Pie charts inside it then pass the touched position to the charts and get the calculated response (check pie chart sample 4).

### DIFF
--- a/example/lib/presentation/samples/chart_samples.dart
+++ b/example/lib/presentation/samples/chart_samples.dart
@@ -22,6 +22,7 @@ import 'line/line_chart_sample9.dart';
 import 'pie/pie_chart_sample1.dart';
 import 'pie/pie_chart_sample2.dart';
 import 'pie/pie_chart_sample3.dart';
+import 'pie/pie_chart_sample4.dart';
 import 'radar/radar_chart_sample1.dart';
 import 'scatter/scatter_chart_sample1.dart';
 import 'scatter/scatter_chart_sample2.dart';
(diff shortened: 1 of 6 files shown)


imaNNeo:
There is a better solution to work with the low-level [RenderBox/RenderObject](https://api.flutter.dev/flutter/rendering/RenderBox-class.html) and the [HitTest topic.](https://api.flutter.dev/flutter/rendering/RenderBox/hitTestSelf.html)

https://github.com/imaNNeo/fl_chart/blob/803409c7c428579f968b276ad9f125ac3c7150d9/lib/src/chart/base/base_chart/render_base_chart.dart#L109

Basically, we need to run `getResponseAtLocation()` and return false if response is null (or something like that, you need to take a look)

https://github.com/imaNNeo/fl_chart/blob/803409c7c428579f968b276ad9f125ac3c7150d9/lib/src/chart/base/base_chart/render_base_chart.dart#L175-L177

btw, if you think it is a complicated topic (which is actually complicated), I can do that.

imaNNeo:
Closing, because it had no activity for a while

nitin-janyani:
I'm looking for the same feature only.

PR #1535, opened 2023-12-26
Upgrade Kotlin version from 1.6.10 to 1.7.10

jerryzhenleicai:
I am using Android Studio Giraffe | 2022.3.1 Patch 2
when building the fl_chart example, I got an error about Kotlin version. So I upgraded Kotlin version in build.gradle and fl_chart builds and runs fine.


imaNNeo:
Thank you!

Issue #1539, opened 2023-12-31
Some lines in bar chart slide to the right, others move vertically 


TimeLord2010:
I'm creating a timeline style of graph using line chart where I update the data based on time every x amount of seconds. The issue is that some lines slide to the right, as expected, but other simply move vertically.

**Reproduce**
```dart
  List<LineChartBarData> _getDatas(WidgetRef ref) {
    List<LineChartBarData> datas = [];
    for (final host in hosts) {
      var provider = ref.watch(pingListProvider(host));
      List<FlSpot> points = provider.points;
      var data = LineChartBarData(
        spots: points,
        isStepLineChart: true,
        show: points.isNotEmpty,
        color: CustomColors.fromString(provider.host),
        dotData: FlDotData(
          show: false,
        ),
      );
      datas.add(data);
    }
    return datas;
  }
```

**Video**
[Video](https://drive.google.com/file/d/1G7ZM6vZFfmXKRvK0E9zVDdtMHTWW50pB/view?usp=sharing)

**To Reproduce**
[Repository](https://gitlab.com/vinivelloso/network_analysis)
- Chart generation code:
```dart
LineChart(
            swapAnimationCurve: Curves.easeInOut,
            swapAnimationDuration: const Duration(milliseconds: 250),
            LineChartData(
              minY: minPing == null ? 0 : minPing - 5,
              maxY: maxPing == null ? 0 : maxPing * 1.05,
              gridData: FlGridData(
                horizontalInterval: 25,
                verticalInterval: 5, // 5 seconds
              ),
              titlesData: _getTitlesData(),
              lineBarsData: _getDatas(ref),
              borderData: FlBorderData(
                show: false,
                border: Border.all(
                  color: Colors.grey.shade200,
                ),
              ),
            )
```

**Notes**
Strangely, the lines that are above 100 all move vertically, and the lines below 100 slide.

**Versions**
 - Flutter 3.16.4
 -  fl_chart: ^0.62.0

imaNNeo:
It is because of an animation that we have inside the chart. 
So when a list changes and the lengths is the same, we lerp through all the elements and animate them. So in this case we should not do that.
For now, you can disable the animations by setting zero in the `swapAnimationDuration`.

If you or someone else want to work on it, this is a good starting point:
https://github.com/imaNNeo/fl_chart/blob/d1166014244e03df3c38ff8f55fa9e78c4740ecb/lib/src/utils/lerp.dart#L6-L24

PR #1556, opened 2024-01-14
feat: Added flag to disable vertical animations on line chart

TimeLord2010:
In response to the issue #1539.


TimeLord2010:
Sorry... Seems that this doesn't always work.

PR #1560, opened 2024-01-22
Bugfix/memory leak

imaNNeo:


lib/src/chart/bar_chart/bar_chart_data.dart
```diff
@@ -10,6 +11,9 @@ import 'package:fl_chart/src/utils/lerp.dart';
 import 'package:fl_chart/src/utils/utils.dart';
 import 'package:flutter/material.dart';
 
+final _isTest = Platform.environment.containsKey('FLUTTER_TEST');
```
NotTsunami, line 14: @imaNNeo This line is causing build failures for web (#1565). There is a PR to fix this that has several upvotes (#1568). Would be nice to see CI for all platforms so we can avoid accidentally breaking a platform in the future.

imaNNeo: Thanks for letting me know!

imaNNeo:
Thanks!

PR #1578, opened 2024-02-03
Declare supported platforms explicitly

### LINKED ISSUE #1577: Web missing from supported platforms on pub.dev
imagem

Since **fl_chart** is fully functional on the web it should be shown as a supported platform. The Dart [docs](https://dart.dev/tools/pub/pubspec#platforms) explain that the supported list is auto-detected and should be manually set in the `pubspec.yaml` if it is incorrect.

It seems this was already previously discussed and fixed, in #780, but the issue has returned. Setting it manually would prevent the same from ever happening again.

The missing Web support label could turn away newcomers since there is not a lot of mention in any of the package pages about the platforms it works on.

It also made me personally have to double check if the package was actually supported on the Web, since I experienced the issue mentioned on #1565 and didn't know if it was intended behavior.

This would make it clear that all current platforms are supported and that failure to work in any of them is indeed a bug and not intended.

dy0gu:
Fixes #1577

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -50,3 +50,11 @@ topics:
   - graph
   - diagram
 
+platforms:
+  android:
+  ios:
+  linux:
+  macos:
+  web:
+  windows:
+


stwarwas:
I don't think this fixes the issue since the code makes use of dart.io (see ticket).

imaNNeo:
Yes, I think it doesn't help too

PR #1579, opened 2024-02-04
 Vertical text for FLineLabel

### LINKED ISSUE #1574: How can I show a label VerticalLineLabel in vertical ?
Hello, i want to show a 'Accumulation' label from VerticalLineLabel in vertical as the image below. Could you please give me some ideals ?

Screenshot 2024-01-30 at 11 11 19

julien4215:
Closes #1574.

The goal of this PR is give the option to display the text of the label vertically for the `VerticalLineLabel` and `HorizontalLineLabel` (see the screenshot in the issue #1574).


imaNNeo:
With `alignment: Alignment.topLeft and direction: LabelDirection.vertical`, it draws outside of the chart
image

But with `alignment: Alignment.topRight and direction: LabelDirection.vertical` it draws inside of the chart.
So I think we should draw it inside in either ways.
image

imaNNeo:
Please write some unit-tests

julien4215:
> One question, does it make sense to have `angle` instead of limited direction values? With angle, I think we are free to do anything. But I agree that it is more difficult and challenging to implement.

It makes sense to have an angle parameter as the label could be displayed in each direction, however it would require a lot of effort for really rare use cases.

imaNNeo:
> > One question, does it make sense to have `angle` instead of limited direction values? With angle, I think we are free to do anything. But I agree that it is more difficult and challenging to implement.
> 
> It makes sense to have an angle parameter as the label could be displayed in each direction, however it would require a lot of effort for really rare use cases.

You're right. Okay so let's stick to the current implementation and write some unit-tests please

julien4215:
> Please write some unit-tests

Done

julien4215:
> With `alignment: Alignment.topLeft and direction: LabelDirection.vertical`, it draws outside of the chart 
> But with `alignment: Alignment.topRight and direction: LabelDirection.vertical` it draws inside of the chart. So I think we should draw it inside in either ways.
This behavior is the same for `VerticalLineLabel` so it makes sense for me to keep the same logic.

With `alignment: Alignment.bottomLeft` and direction: LabelDirection.horizontal`, it draws outside of the chart 

With `alignment: Alignment.topLeft` and direction: LabelDirection.horizontal`, it draws inside of the chart

imaNNeo:
> > With `alignment: Alignment.topLeft and direction: LabelDirection.vertical`, it draws outside of the chart
> > But with `alignment: Alignment.topRight and direction: LabelDirection.vertical` it draws inside of the chart. So I think we should draw it inside in either ways.
> 
> This behavior is the same for `VerticalLineLabel` with `direction: LabelDirection.horizontal` so it makes sense for me to keep the same logic.
> 
> With `alignment: Alignment.bottomLeft` and direction: LabelDirection.horizontal`, it draws outside of the chart 
> 
> With `alignment: Alignment.topLeft` and direction: LabelDirection.horizontal`, it draws inside of the chart 

I think the vertical one is a bug too. I mean we need to fix both of them. 
It would be nice if you can fix them here, otherwise I will do this in a separate PR.
What do you think?

julien4215:
> I think the vertical one is a bug too. I mean we need to fix both of them.
> It would be nice if you can fix them here, otherwise I will do this in a separate PR.
> What do you think ?

I agree that it would be better like this. I'll add a commit to this PR for both of them.

imaNNeo:
> I agree that it would be better like this. I'll add a commit to this PR for both of them.

That would be wonderful

julien4215:
> I agree that it would be better like this. I'll add a commit to this PR for both of them.
Done

lib/src/chart/base/axis_chart/axis_chart_data.dart
```diff
@@ -1149,9 +1146,9 @@ class HorizontalLineLabel extends FlLineLabel with EquatableMixin {
   ) {
     return HorizontalLineLabel(
       padding:
-          EdgeInsets.lerp(a.padding as EdgeInsets, b.padding as EdgeInsets, t),
+          EdgeInsets.lerp(a.padding as EdgeInsets, b.padding as EdgeInsets, t)!,
       style: TextStyle.lerp(a.style, b.style, t),
-      alignment: Alignment.lerp(a.alignment, b.alignment, t),
+      alignment: Alignment.lerp(a.alignment, b.alignment, t)!,
       labelResolver: b.labelResolver,
```
imaNNeo, line 1152: Please add the direction property here in the lerp method

lib/src/chart/base/axis_chart/axis_chart_data.dart
```diff
@@ -1211,9 +1203,9 @@ class VerticalLineLabel extends FlLineLabel with EquatableMixin {
   ) {
     return VerticalLineLabel(
       padding:
-          EdgeInsets.lerp(a.padding as EdgeInsets, b.padding as EdgeInsets, t),
+          EdgeInsets.lerp(a.padding as EdgeInsets, b.padding as EdgeInsets, t)!,
       style: TextStyle.lerp(a.style, b.style, t),
-      alignment: Alignment.lerp(a.alignment, b.alignment, t),
+      alignment: Alignment.lerp(a.alignment, b.alignment, t)!,
       labelResolver: b.labelResolver,
```
imaNNeo, line 1209: Also here

lib/src/chart/base/axis_chart/axis_chart_data.dart
```diff
@@ -1176,25 +1174,19 @@ class VerticalLineLabel extends FlLineLabel with EquatableMixin {
...
-        );
+    super.padding = const EdgeInsets.all(6),
+    super.style = const TextStyle(
+      color: Colors.black,
+      fontWeight: FontWeight.bold,
+      fontSize: 14,
+    ),
+    super.alignment = Alignment.bottomRight,
+    super.show = false,
+    super.direction = LabelDirection.horizontal,
```
imaNNeo, line 1187: Does it make sense to have a vertical direction by default here?

imaNNeo: I realized that the default behavior is horizontal. So we don't need to change it. 
Thanks for keeping it the same as before.

imaNNeo:
It's ready to be merged after those minor changes.
Thanks in advance!

lib/src/chart/base/axis_chart/axis_chart_data.dart
julien4215: I think the label direction doesn't matter for the linear interpolation. Am I wrong ?

julien4215: I think the label direction doesn't matter for the lerp method. Am I wrong ?

imaNNeo: For the boolean types, we use the destination value directly. So we can say
```
direction: b.direction,
```

julien4215: I am not sure about what you mean by 

> For the boolean types, we use the destination value directly

but I added the direction property in the lerp method

PR #1580, opened 2024-02-06
Removed dependency to dart:io in order to support Web #1577

stwarwas:
This PR is related to issue #1577.

Two classes used dart:io to determine whether the code is being executed as part of a test.

This prevented the usage of fl_chart for the Web as the target platform.

I moved the test-related initialization code from the production code to the unit tests.

### DIFF
--- a/lib/src/chart/bar_chart/bar_chart_data.dart
+++ b/lib/src/chart/bar_chart/bar_chart_data.dart
@@ -1,19 +1,13 @@
 // coverage:ignore-file
-import 'dart:io';
 import 'dart:math';
 import 'dart:ui';
 
 import 'package:equatable/equatable.dart';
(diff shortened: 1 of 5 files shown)


NotTsunami:
@stwarwas You need to rebase this

@imaNNeo This refactor is probably the correct way of solving the issue at hand

imaNNeo:
Oh!
I didn't know if we use dart.io, pub doesn't suggest it as a web package.
Can you please rebase your branch?

stwarwas:
@imaNNeo I fixed the conflicts. Can you use it like this or do I have to do something else?

imaNNeo:
You need to rebase your branch (instead of `merge`)

imaNNeo:
Thanks!

Issue #1593, opened 2024-02-22
line chart shows negative y-axis, but there are no negative numbers


williamluotao:
** Don't make a duplicate issue.
You can search in issues to make sure there isn't any already opened issue with your concern.

**Describe the bug**

**To Reproduce**
Provide us a completely reproducible code (contains the main function) in a `main.dart` file, it helps us to find the bug immediately.
<img width="338" alt="Screen Shot 2024-02-22 at 19 05 11" src="https://github.com/imaNNeo/fl_chart/assets/1224106/2105e7e2-d3dd-40d3-baff-91603df2ba82">

**Screenshots**
If applicable, add screenshots, or videoshots to help explain your problem.

**Versions**
 - which version of the Flutter are you using?
 - which version of the FlChart are you using?

harshpjoshi:
@williamluotao I able to fixing this issue by adding some Path logic. Here is the PR:[1594](https://github.com/imaNNeo/fl_chart/pull/1594)
@imaNNeo Please check if it is a valid fix for same.
Thanks

imaNNeo:
There is a solution at the moment, you can set `preventCurveOverShooting: true` in your LineChartBarData.
But I will take a look at the @harshpjoshi's solution.
Please provide me a reproducible code.

williamluotao:
> There is a solution at the moment, you can set `preventCurveOverShooting: true` in your LineChartBarData. But I will take a look at the @harshpjoshi's solution. Please provide me a reproducible code.

LineChartData _createTimeSequenceData(List dt_list) {
      print("dt_list ${dt_list}");
      List<FlSpot> spots = [];
      List<int> showingTooltipOnSpots = [];

      List<Map<String, dynamic>> b = List.generate(
        24,
            (index) => {"recHour": index, "times": 0},
      );
      List<double> stops = [];

      for (ChartSequence cs in dt_list) {
        showingTooltipOnSpots.add(cs.times);
        stops.add((cs.times).toDouble());

        var recHour = cs.recHour;
        var times = cs.times;
        var index = b.indexWhere((element) =>
        element["recHour"] ==
            int.parse(recHour.replaceAll(RegExp(r'[^0-9]'), '')));

        if (index != -1) {
          b[index]["times"] = times;
        }
      }
      //print(b);

      for (var entry in b) {
        spots.add(
            FlSpot(entry["recHour"].toDouble(), entry["times"].toDouble()));
      }
      //print(spots);


[...]

PR #1594, opened 2024-02-27
Fixed line chart shows negative values

harshpjoshi:
fix: Adjust bar path to fit within a specified rectangular area, #1593

This commit adds a new function called `adjustPath` that takes a `Path` and a `Rect` as parameters. The function adjusts the given path to fit within the specified rectangular area by clamping the path's points to the rect's boundaries. This ensures that the bar path is always fully visible within the chart area.

The function iterates over the path metrics of the input path and for each path metric, it iterates over its length. For each iteration, it gets the tangent at the current offset and checks if the tangent's position is within the rectangular area. If it's not, the position is clamped to the rect's boundaries. Finally, the adjusted path is returned.

### DIFF
--- a/lib/src/chart/line_chart/line_chart_painter.dart
+++ b/lib/src/chart/line_chart/line_chart_painter.dart
@@ -974,6 +974,7 @@ class LineChartPainter extends AxisChartPainter<LineChartData> {
       ..transparentIfWidthIsZero();
 
     barPath = barPath.toDashedPath(barData.dashArray);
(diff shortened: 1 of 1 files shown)


imaNNeo:
I think this approach adds some bugs. For example, there might be some displacements between the line and the grid lines behind it.

imaNNeo:
Ahh, now I see the issue.

imaNNeo:
Closed because of inactivity

PR #1636, opened 2024-04-17
Add curved step line edges

mariogrip:
This adds curved step line edges, putting this as a draft to get some feedback. 

bilde


imaNNeo:
Hi, thanks for your contribution. I'm sorry about my late reply.
It looks good to me, 
I think we can remove `isStepLineEdgesCurved` property, as we can having it using our `stepLineEdgesSmoothness`. It can be `stepLineEdgesSmoothness != 0`

I think you need to rebase your branch.

Also please update the bar_chart.md file to add this property,

PR #1666, opened 2024-05-15
Add rangeAnnotations to ScatterCharts

### LINKED ISSUE #1665: ScatterChart doesn't support rangeAnnotation.
I would like to use a scatter plot, but I need the range annotations to color part of the background of the chart.  This is easy to do using `LineChart`, but not for `ScatterChart` even though the `rangeAnnotations` property is in a common ancestor.

The ability to set this property needs to be exposed in the `ScatterChartData` through the constructor just as it is exposed in the `LineChartData` constructor.

derrickgw:
Although `ScatterChartData` inherits from `AxisChartData`, `rangeAnnocations` was not accessible though the constructions.  

This exposes that property the same way `LineChartData` does.

Fixes #1665


imaNNeo:
Hi, thanks for your contribution and I'm very sorry about the late reply.
It looks good to me. Can you please rebase your branch and fix the pipeline issues?

imaNNeo:
Closed because of inactivity

PR #1669, opened 2024-05-19
add gradient color in Radarchart

sudhar08:
This issue proposes enhancing the RadarChart functionality by adding support for ability customization and color filling options. The goal is to provide users with more control over the appearance and functionality of their RadarCharts, enabling better data visualization and improved user experience.


imaNNeo:
Hi. As I see you just defined the `gradientcolor` and never used it in the drawing functions.
Also, it has a typo, it should be `gradientColor`

PR #1682, opened 2024-06-14
Add isMinValueAtCenter property to RadarChartData

### LINKED ISSUE #1351: Radar plot center point is zero
image
I want the points with the value of zero to be concentrated on the center point as shown in the figure, but I don't know how to set it

### LINKED ISSUE #1442: How can I set the center point to 0, fix the maximum value to 100, and not require my data to include the maximum value?
Radar chart, how can I fix the maximum value to 100, the minimum value to 0 (center point), even if my dataEntries may not include 100. Currently, I noticed that the scale adjusts based on the maximum value in dataEntries, but my values may not always include the maximum value (100). How should I handle this?

`                RadarDataSet(
                  // 设置数据内部的颜色
                  fillColor: ColorStyle.system40_0054a7,
                  // 设置数据边框的颜色
                  borderColor: ColorStyle.system_0054a7,
                  //设置数据边框的宽度
                  borderWidth: 2,
                  //设置数据点的半径(点的大小)为0来隐藏它们
                  entryRadius: 0,
                  dataEntries:  [
                    RadarEntry(value:0),
                    RadarEntry(value: 0),
                    RadarEntry(value:0),
                    RadarEntry(value: 90),
                    RadarEntry(value: 70),
                    RadarEntry(value: 50),
                  ]
                ),`

image

**Versions**
 - flutter sdk: '>=3.0.6 <4.0.0'
 -  fl_chart: ^0.63.0

soraef:
### Summary
This PR adds the `isMinValueAtCenter` flag to the `RadarChartPainter` class to allow the minimum value to be centered in the RadarChart.

### Related Issues
- Fixes #1351 
- Fixes #1442

### Changes
- Added `isMinValueAtCenter` flag to `RadarChartData`
- Updated `RadarChartPainter` methods to support centering the minimum value when `isMinValueAtCenter` is true

### isMinValueAtCenter Comparison

<table>
  <tr>
    <td><strong>isMinValueAtCenter is false</strong></td>
    <td><strong>isMinValueAtCenter is true</strong></td>
  </tr>
  <tr>
    <td>isMinValueAtCenter is false</td>
    <td>isMinValueAtCenter is true</td>
  </tr>
</table>

### DIFF
--- a/lib/src/chart/radar_chart/radar_chart_data.dart
+++ b/lib/src/chart/radar_chart/radar_chart_data.dart
@@ -203,6 +203,7 @@ class RadarChartData extends BaseChartData with EquatableMixin {
     BorderSide? tickBorderData,
     BorderSide? gridBorderData,
     RadarTouchData? radarTouchData,
+    bool? isMinValueAtCenter,
     FlBorderData? borderData,
   }) =>
       RadarChartData(
@@ -219,6 +220,7 @@ class RadarChartData extends BaseChartData with EquatableMixin {
         tickBorderData: tickBorderData ?? this.tickBorderData,
         gridBorderData: gridBorderData ?? this.gridBorderData,
         radarTouchData: radarTouchData ?? this.radarTouchData,
+        isMinValueAtCenter: isMinValueAtCenter ?? this.isMinValueAtCenter,
         borderData: borderData ?? this.borderData,
       );
 
@@ -245,6 +247,7 @@ class RadarChartData extends BaseChartData with EquatableMixin {
         radarShape: b.radarShape,
         tickBorderData: BorderSide.lerp(a.tickBorderData, b.tickBorderData, t),
         borderData: FlBorderData.lerp(a.borderData, b.borderData, t),
+        isMinValueAtCenter: b.isMinValueAtCenter,
         radarTouchData: b.radarTouchData,
       );
     } else {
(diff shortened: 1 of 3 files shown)


soraef:
Thank you for the wonderful product. I use it all the time.

After implementing the changes, I realized that the isMinValueAtCenter flag might not be necessary. Should the minimum value be centered by default? If the current behavior is a bug, it would be better to make the minimum value centered by default instead of adding a flag. This would simplify the code and avoid unnecessary complexity.

imaNNeo:
Yes I think it is a bug and it should always be centered

rickypid (approved):

soraef:
I identified a bug where the behavior did not center around zero as expected, so I made corrections. However, I couldn’t get the test to pass. I'm not sure how to adjust the test code to reflect the fix, so I've left it unresolved.

lib/src/chart/radar_chart/radar_chart_data.dart
```diff
@@ -154,6 +155,9 @@ class RadarChartData extends BaseChartData with EquatableMixin {
   /// Handles touch behaviors and responses.
   final RadarTouchData radarTouchData;
 
+  /// If [isMinValueAtCenter] is true, the minimum value of the [RadarChart] will be at the center of the chart.
+  final bool isMinValueAtCenter;
```
imaNNeo, line 159: You need to add this property in the `copyWith()`, `lerp()` and the `List<Object?> get props => []` functions

PR #1698, opened 2024-06-27
Added option to added Rounded Capped PieChart

Begarudev:
enhacement asked in this issue #1700

### DIFF
--- a/lib/src/chart/pie_chart/pie_chart_data.dart
+++ b/lib/src/chart/pie_chart/pie_chart_data.dart
@@ -79,6 +79,9 @@ class PieChartData extends BaseChartData with EquatableMixin {
   /// and replaces provided values.
 
   /// The style of the cap at the end of the pie chart
+  /// It Accepts [CapStyle.none] for no cap, [CapStyle.startCapped] for a cap at the start of the chart,
+  /// [CapStyle.endCapped] for a cap at the end of the chart.
+  /// Default is [CapStyle.none]
   final CapStyle capstyle;
 
   PieChartData copyWith({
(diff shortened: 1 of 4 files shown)


lib/src/chart/pie_chart/pie_chart_data.dart
```diff
@@ -31,6 +31,10 @@ class PieChartData extends BaseChartData with EquatableMixin {
     PieTouchData? pieTouchData,
     FlBorderData? borderData,
     bool? titleSunbeamLayout,
+    CapStyle? capStyle,
+    bool? shadow,
```
imaNNeo, line 35: Isn't it better to use the [Shadow](https://api.flutter.dev/flutter/dart-ui/Shadow-class.html) object?

Something like what we have in the LineChart:
https://github.com/imaNNeo/fl_chart/blob/11d13a78e6a134aa0cfdcd8dfa88dceaf79a4033/lib/src/chart/line_chart/line_chart_data.dart#L360-L361

And this is how we paint it in the `drawBarShadow()` function:
https://github.com/imaNNeo/fl_chart/blob/11d13a78e6a134aa0cfdcd8dfa88dceaf79a4033/lib/src/chart/line_chart/line_chart_painter.dart#L906-L939

imaNNeo: And if it is transparent (default value), we will not draw it.

imaNNeo:
As I see, it also tries to fix #1699.
But I think it would be better if you create separate PRs for this specific case to follow it up.

moritz157:
Any news on this? I would really like to be able to round caps in donut charts too and looking at this it looks like it's pretty much solved and just has to be merged (and maybe reorganized first), right?

imaNNeo:
@moritz157 It seems the PR author is not available anymore. Also, the branch is outdated. So I think we have to close it.

diegoveloper:
any chance to re-open and finish the implementation of this feature? :D

PR #1713, opened 2024-07-17
Calculating side titles and horizontal lines from minY value

ChrisElliotUK:
This is a rough idea and I needed to make it quickly work for myself. Basically I want the intervals to be calculated from teh minY value. Currently you often end up with an awkward interval at the top and bottom of the graph. This way the interval is simply added to the minY value so it is even. It only works well if you calculate the interval yourself. Would love to see someone take this idea and flush it out properly but I don’t have th time atm


imaNNeo:
I think it would be better if you open an issue and explain it, then propose your solution for the issue. Then people can read and contribute to the idea.
So please open and issue to do that, and remove the PR afterward.

Cheers!

PR #1721, opened 2024-08-03
added detectScale to FlTouchData flag and scale gesture detection option

eladm-ultrawis:
added detectScale to FlTouchData flag and scale gesture detection option in the touch callback to all charts.

enabling scale detection and pan gesture detection is not supported, so enabling scale detection will disable pan detection and vice versa

related issues:
This PR addresses the problem described in issue #1720 and #71

Issue #1720, opened 2024-07-27
Feature Request - add support for scale gesture recognition for LineTouchData.touchCallback


eladm-ultrawis:
I thoroughly investigated a solution for supporting scale and panning of a line plot by manipulating minX and minY. I mostly followed the suggestions offered in this open [issue](https://github.com/imaNNeo/fl_chart/issues/71).

After trying many of the offered solutions and a few of my own I reached the following conclusions:
1. supporting panning of a plot is best done using `LineTouchData.touchCallback` and handling `FlPanUpdateEvent`. 
2. supporting zoom is very hard to do: using a `GestureDetector` / `Listener` only seems to work properly if my FlChart is warpped in an `IgnorePointer` widget. and unfortunately `LineTouchData.touchCallback` does not support scale events.

I forked fl_chart and tried to add a `ScaleGestureRecognizer` to `RenderBaseChart` and added a bunch of `Fl***Events` for scale events. I followed the usage pattern for the other gesture recognizers however this doesn't work.

Could you advice on how this should be added?

Here is the essence of what I tried to add in `render_base_chart.dart`:

```
  void initGestureRecognizers() {
    _scaleGestureRecognizer = ScaleGestureRecognizer()
      ..onStart = (details) {
        _notifyTouchEvent(FlScaleStartEvent(details));
      }
      ..onUpdate = (details) {
        _notifyTouchEvent(FlScaleUpdateEvent(details));
      }
      ..onEnd = (details) {
        _notifyTouchEvent(FlScaleEndEvent(details));
      };
```

and here:

```
  @override
  void handleEvent(PointerEvent event, covariant BoxHitTestEntry entry) {
    assert(debugHandleEvent(event, entry));
    if (_touchCallback == null) {
      return;
    }
    if (event is PointerDownEvent) {
      _longPressGestureRecognizer.addPointer(event);
      _tapGestureRecognizer.addPointer(event);
      _panGestureRecognizer.addPointer(event);
      _scaleGestureRecognizer.addPointer(event);
    } else if (event is PointerHoverEvent) {
      _notifyTouchEvent(FlPointerHoverEvent(event));
    }
  }
```

eladm-ultrawis:
Ok, Apparently the `ScaleGestureRecognizer` starts working if it replaces the `PanGestureRecognizer` entirely. 
[...]

PR #1728, opened 2024-08-15
Both scaleable and draggable chart implemented

adamsocrat:
Implements #1720 

I used the idea from #1721 and implemented FlScaleEvent but without explicitly specifying `detectScale` to disable drag fully when scaling. `detectScale` could've been set dynamically but it was combursome and couldn't get it working. Implementing a new FLTouchEvent seemed better option.

The drawback was as in #1721 mentioned `PanGestureRecognizer ` and `ScaleGestureRecognizer` does not like to be used together. But `ScaleGestureRecognizer` event details also feeds the `PanGestureRecognizer` event details (except onDown) [see](https://stackoverflow.com/a/66122623/12263240), so instead of disabling dragging fully when you want to scale, I notify ScaleEvent when there are two points and drag when one point is in contact with the screen. Same idea of this [comment](https://github.com/imaNNeo/fl_chart/issues/1720#issuecomment-2268291918) 

**! Details feeding into `FlPan...Event` have been switched to Scale details. Testing may be needed.**

You can get TouchEvent from touchcallback as 
```dart
touchCallback: (FlTouchEvent touchEvent, lineTouchRsp) {
              if (touchEvent is FlScaleStartEvent) {
                       // https://api.flutter.dev/flutter/gestures/ScaleStartDetails-class.html
                       }
             if (touchEvent is FlScaleUpdateEvent) {
                      // You can see the details of touch event with:  touchEvent.details
                     // https://api.flutter.dev/flutter/gestures/ScaleUpdateDetails-class.html
                       }
             if (touchEvent is FlPanUpdateEvent) {
                     // NOT https://api.flutter.dev/flutter/gestures/DragUpdateDetails-class.html 
                     // https://api.flutter.dev/flutter/gestures/ScaleUpdateDetails-class.html
                       }
}
```

with the `FlScaleEvent` I am able to set `minX` and `maxX` to zoom in/out chart without the scale is being also interpreted as drag. See #71


TheOlajos:
Would be great if this could be merged for the next implementation.

lib/src/chart/base/base_chart/render_base_chart.dart
```diff
@@ -162,6 +167,23 @@ abstract class RenderBaseChart<R extends BaseTouchResponse> extends RenderBox
...
+  void _notifyScaleEvent(FlTouchEvent event) {
+    if (_touchCallback == null) {
+      return;
+    }
+
+    final pointerCount = event.pointerCount;
+
+    R? response;
+    if (pointerCount > 1) {
+      _touchCallback!(event, response);
```
imaNNeo, line 183: You can just set a null as the second parameter (instead of having a nullable R?, which is always null)

lib/src/chart/base/base_chart/render_base_chart.dart
```diff
@@ -50,22 +52,20 @@ abstract class RenderBaseChart<R extends BaseTouchResponse> extends RenderBox
...
   void initGestureRecognizers() {
-    _panGestureRecognizer = PanGestureRecognizer();
-    _panGestureRecognizer
-      ..onDown = (dragDownDetails) {
-        _notifyTouchEvent(FlPanDownEvent(dragDownDetails));
+    _scaleGestureRecognizer = ScaleGestureRecognizer();
+    _scaleGestureRecognizer
+      ..onStart = (details) {
+        _notifyScaleEvent(FlScaleStartEvent(details));
+        _notifyTouchEvent(FlPanDownEvent(details));
```
imaNNeo, line 59: Is it okay to call both FlPanDown and FlPanStart at the same time?

lib/src/chart/base/base_chart/render_base_chart.dart
```diff
@@ -50,22 +52,20 @@ abstract class RenderBaseChart<R extends BaseTouchResponse> extends RenderBox
...
   /// Initializes our recognizers and implement their callbacks.
   void initGestureRecognizers() {
-    _panGestureRecognizer = PanGestureRecognizer();
-    _panGestureRecognizer
-      ..onDown = (dragDownDetails) {
-        _notifyTouchEvent(FlPanDownEvent(dragDownDetails));
+    _scaleGestureRecognizer = ScaleGestureRecognizer();
+    _scaleGestureRecognizer
+      ..onStart = (details) {
+        _notifyScaleEvent(FlScaleStartEvent(details));
```
imaNNeo, line 58: Does it get called for the second finger (when you already have your first finger, and try to add the second finger to do the scale gesture)?

imaNNeo: In this case, I think we should have something like fingerIndex or something like that

lib/src/chart/base/base_chart/render_base_chart.dart
```diff
@@ -50,22 +52,20 @@ abstract class RenderBaseChart<R extends BaseTouchResponse> extends RenderBox
...
+        _notifyTouchEvent(FlPanDownEvent(details));
+        _notifyTouchEvent(FlPanStartEvent(details));
       }
-      ..onStart = (dragStartDetails) {
-        _notifyTouchEvent(FlPanStartEvent(dragStartDetails));
+      ..onUpdate = (details) {
+        _notifyScaleEvent(FlScaleUpdateEvent(details));
+        _notifyTouchEvent(FlPanUpdateEvent(details));
```
imaNNeo, line 64: And also for here, (something like fingerIndex)

eladm-ultrawis:
Hi @adamsocrat  @imaNNeo , joining the conversation here.

I think it might be a bit inconsistent with the rest of the flutter framework to raise scale events for 2 fingers and pan for 1 finger.

To my limited understanding, the idea behind this is that - scale is also a pan. with the scale events, you could implement pan because it already has all the necessary information. 
I think whatever implementation you guys end choosing, should be consistent with how GestureDetector works in this scenario.

lib/src/chart/base/base_chart/render_base_chart.dart
adamsocrat: Will fix.

lib/src/chart/base/base_chart/render_base_chart.dart
adamsocrat: Probably not, drag down details and scalestart details will be different and will be breaking a change. Thinking of another solution

adamsocrat:
> Hello, any progress with this ?

Unfortunately I had no time to debug and implement better. I think in #1793 they are working on a better solution. So I am closing this issue.

PR #1732, opened 2024-08-25
Pie Chart Connector Lines feature

### LINKED ISSUE #1731: Need arrow line to pie chart section title
We can have a flag which will show arrow line to the pie chart section title. Please check the attached image.

image

amirk3321:
Hello,

I added a new feature to `fl_chart` package. Now, developers can use `FLConnectorLineSettings` to configure line pointing the value for `PieChartSectionData`.

Example `PieChartSectionData` to draw connector line pointing the value:

```dart
PieChartSectionData(
  connectorLineSettings: const FLConnectorLineSettings(
          type: FLConnectorType.line,
          color: Color(0xFFD5E5EC), // use any color,
          length: "40%" // use any length
  ),
);
```
The line results:

Screenshot 2024-08-25 at 12 47 04 PM

The `connectorLineSettings` field in `PieChartSectionData` is nullable, so simply remove it if you don't want to use it.

The suggestions, bugs, or contribution to the enhancements of these lines would be greatly appreciated.

Thanks

Fixes #1731


cogivn:
Hi @amirk3321, I also need this feature on my app. <br/> Can we add more customization like ability to customize external labels? Something like this...
Thank you so much,

lib/src/chart/pie_chart/pie_chart_connector_lines.dart
```diff
@@ -0,0 +1,18 @@
+import 'package:flutter/material.dart';
+
+/// Type of connector line.
+enum FLConnectorType { line }
```
imaNNeo, line 4: I didn't understand the use case of this enum
We have FLConnectorLineSettings (which is a line itself as the name suggests), why we need a FLConnectorType (which is again a line)?

amirk3321: I added this to keep an idea for those who wanted enhance this more, for example, might be another enum `.dotted`, `.dashed` and so on. I'll remove it in the next commit to clean it make it ready for merge. Thanks for highlighting.

lib/src/chart/pie_chart/pie_chart_data.dart
```diff
@@ -112,7 +113,7 @@ class PieChartData extends BaseChartData with EquatableMixin {
         pieTouchData: b.pieTouchData,
         sectionsSpace: lerpDouble(a.sectionsSpace, b.sectionsSpace, t),
         startDegreeOffset:
-            lerpDouble(a.startDegreeOffset, b.startDegreeOffset, t),
+        lerpDouble(a.startDegreeOffset, b.startDegreeOffset, t),
```
imaNNeo, line 116: And here

lib/src/chart/pie_chart/pie_chart_data.dart
```diff
@@ -309,33 +313,28 @@ class PieTouchData extends FlTouchData<PieTouchResponse> with EquatableMixin {
     MouseCursorResolver<PieTouchResponse>? mouseCursorResolver,
     Duration? longPressDuration,
   }) : super(
-          enabled ?? true,
-          touchCallback,
-          mouseCursorResolver,
-          longPressDuration,
-        );
+    enabled ?? true,
```
imaNNeo, line 316: And here

imaNNeo:
These changes caused some conflicts in your branch. Please revert all the formatting changes

Issue #1715, opened 2024-07-22
Border Radius for each corner in BarTouchTooltipData


cnschnei-btc:
**Is your feature request relasted to a problem? Please describe.**
My Tooltip for the BarChart should have a custom rounded radius. With the current implementation i can only set a border radius for all 4 corners. I need the bottom left corner not to be rounded (like a speechbubble).

**Describe the solution you'd like**
It could be implemented that you can provide a normal BorderRadius.

imaNNeo:
Added the new `final BorderRadius tooltipBorderRadius;` in [0.71.0](https://github.com/imaNNeo/fl_chart/releases/tag/0.71.0).

PR #1754, opened 2024-09-24
custom border radius in BarTouchTooltipData

MattiaPispisa:
**IMPROVEMENT** (by @MattiaPispisa) custom border radius in `BarTouchTooltipData`, `LineTouchTooltipData`, `ScatterTouchTooltipData` #1715

### DIFF
--- a/lib/src/chart/bar_chart/bar_chart_data.dart
+++ b/lib/src/chart/bar_chart/bar_chart_data.dart
@@ -714,7 +714,8 @@ class BarTouchTooltipData with EquatableMixin {
   /// [BarChart] shows a tooltip popup on top of rods automatically when touch happens,
   /// otherwise you can show it manually using [BarChartGroupData.showingTooltipIndicators].
   /// Tooltip shows on top of rods, with [getTooltipColor] as a background color,
-  /// and you can set corner radius using [tooltipRoundedRadius].
+  /// and you can set corner radius using [tooltipRoundedRadius], of if you need a custom
+  /// border you can use [tooltipBorderRadius].
   /// If you want to have a padding inside the tooltip, fill [tooltipPadding],
   /// or If you want to have a bottom margin, set [tooltipMargin].
   /// Content of the tooltip will provide using [getTooltipItem] callback, you can override it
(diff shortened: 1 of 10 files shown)


lib/src/chart/bar_chart/bar_chart_data.dart
```diff
@@ -702,6 +705,7 @@ class BarTouchTooltipData with EquatableMixin {
   /// also you can set [fitInsideVertically] true to force it to shift inside the chart vertically.
   BarTouchTooltipData({
     double? tooltipRoundedRadius,
+    BorderRadius? tooltipBorderRadius,
```
imaNNeo, line 708: We already have the `tooltipRoundedRadius` property. If you want to add it, please remove the old one.

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## UNRELEASED
+* **IMPROVEMENT** (by @MattiaPispisa) custom border radius in BarTouchTooltipData, #1715
```
imaNNeo, line 2: You don't need to write a changelog here, please read our [contributing guideline](https://github.com/imaNNeo/fl_chart/blob/main/CONTRIBUTING.md)

MattiaPispisa: I had read the guidelines, and it was recommended to start from an issue. From the ‘Mention the related issues’ section, what I did seemed correct. What should I have done differently?
Thank you.

imaNNeo: Ahh, you're right. I'm sorry I forgot to update one part of the guideline.
I just updated it [here](https://github.com/imaNNeo/fl_chart/pull/1775/commits/cef2d74363c4ddda20ab22c075ead9b86fd75ad0).

As we had a lot of conflicts in the change log part, I decided to update it by myself. So you don't need to write something.

Maybe, it would be a good idea to just copy-paste yours in the PR's description. This way I can just add it.

lib/src/chart/bar_chart/bar_chart_data.dart
imaNNeo: You're right, we should keep it backward-compatible.
So can you please make the old one `tooltipRoundedRadius` deprecated? This way we can remove it in the next releases

MattiaPispisa:
Hi @imaNNeo, no problem.  

I removed the changelog and included what I had written in the issue description. 
I also added the annotation (deprecated) to the parameter.

imaNNeo:
Very nice!
Can you please add the same feature for our LineChart and ScatterChart tooltips to keep the consistency?

Also please rebase your branch.

MattiaPispisa:
Thank you! 

I've applied the same changes to both `LineChart` and `ScatterChart`. The implementation in `LineChart` is slightly different because the constructor is a `const`, and I wanted to keep it that way to avoid introducing a breaking change. The exposed interface remains the same.

I also made a rebased with `main` branch.

imaNNeo:
Everything seems good to me
Thanks for your contribution!
Hope to see more from you!

PR #1782, opened 2024-11-24
Update bar_chart_sample1.dart

yassinedevop:
The parameter `duration` for `BarChart` isn't defined, so  `duration` has been changed to `swapAnimationDuration`.


imaNNeo:
`swapAnimationDuration` is deprecated (since [0.69.2](https://github.com/imaNNeo/fl_chart/releases/tag/0.69.2)). So you should use `duration` from now on.

PR #1789, opened 2024-11-26
fix!: PieChartSectionData equatable

Peetee06:
This PR adds tests for lerp utilities and increases the test coverage from 87.52% to 100%. 

In the process I added `EquatableMixin`, because `PieChartSectionData` with the same `sections` were not considered equal.

As far as tests go, this is a breaking change, because one test case assumed pie chart data to be unequal when their `sections` don't reference the same object, even though the `sections` have the same list elements. 

Please let me know, what you think is the best approach for this for fl_chart.

### DIFF
--- a/test/chart/pie_chart/pie_chart_data_test.dart
+++ b/test/chart/pie_chart/pie_chart_data_test.dart
@@ -73,7 +73,7 @@ void main() {
                 PieChartSectionData(value: 22, color: Colors.green),
               ],
             ),
-        false,
+        true,
       );
 
       expect(
--- a/lib/src/chart/pie_chart/pie_chart_data.dart
+++ b/lib/src/chart/pie_chart/pie_chart_data.dart
@@ -134,7 +134,7 @@ class PieChartData extends BaseChartData with EquatableMixin {
 }
 
 /// Holds data related to drawing each [PieChart] section.
-class PieChartSectionData {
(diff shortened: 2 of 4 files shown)


test/chart/pie_chart/pie_chart_data_test.dart
```diff
@@ -73,7 +73,7 @@ void main() {
                 PieChartSectionData(value: 22, color: Colors.green),
               ],
             ),
-        false,
+        true,
```
Peetee06, line 76: This case expected the `PieChartSectionData`s to be unequal, even though their `sections` contain exactly the same elements in the same order. Technically this could be considered a breaking change if we merge it.

Peetee06: I saw that in the Flutter Framework GitHub any change that breaks a test is considered a "breaking change". 

It might break functionality that expect `section` to be unequal when they don't have the same reference. 

I think it's very unlikely that this will actually happen for anyone. But communicating this as a breaking change ensures trust in consumers of `fl_chat`.

imaNNeo: Makes sense. Thanks for mentioning that. 
So let's consider it as a breaking change.

imaNNeo:
Thanks for contributing!
Your PRs are robust.
I appreciate it.

PR #1793, opened 2024-11-27
feat!: Add Chart zoom and pan

peter-trost:
This PR adds zooming and panning for Charts.

Solves #71

BREAKING CHANGE: 
`BarChart` is not `const` anymore, because we are asserting that start, end and center `BarChartAlignment` can only be used with `ScaleAxis` none or vertical. This is because with these alignments, the bars' positions are not calculated based on available width, so scaling horizontally does not do anything. If we want to allow horizontal scaling for `BarChart`, the painters get more complex, because we have to prevent vertical lines from spacing out horizontally only for this case.

# Tasks
- [x] Allow enabling and disabling zoom and pan
- [x] Allow locking pan to horizontal or vertical axis, additional to allowing free panning
- [x] Allow zooming via mouse wheel and trackpad scroll
- [x] Hide title and touch tooltips as soon as corresponding chart point leaves chart bounds
- [x] Add functionality to BarChart
- [x] Add functionality to ScatterChart
- [x] Tests for LineChart + renderer + painter
- [x] Tests for BarChart + renderer + painter
- [x] Tests for ScatterChart + renderer + painter
- [x] Additional tests
- [x] fix LineChart sample 12 overflows on small devices

# Demo
## `LineChart`

https://github.com/user-attachments/assets/b20ae612-344d-43de-9af0-d1eb28ac2a21

## `BarChart`

https://github.com/user-attachments/assets/32eec1dd-1c99-4e01-813c-d01c33c5461e

## `ScatterChart`

https://github.com/user-attachments/assets/4b5832e5-6467-4e46-87c3-22c38d917515


Peetee06:
The current approach I'm trying is using an `InteractiveViewer` with a `TransformationController`. On changes to the `Matrix4` of the controller, I update the data's minX, maxX, minY and maxY. 

Issues I'm facing:
# The axis titles are not scaled correctly yet
Will find a scaling factor that fixes this.

https://github.com/user-attachments/assets/5a5e9be2-5795-450f-ba4e-f1d6ed13901d


# `Clip.hardEdge` clips tooltips
Screenshot 2024-11-27 at 19 47 25

# `Clip.none` allows the graph to overflow
Screenshot 2024-11-27 at 19 46 33

Will look into a way to add a new set of min/max values that constrain the chart's renderer like @imaNNeo suggested.

imaNNeo:
That was a good start,
But just to mention that sometimes we want to limit the zooming/panning only for one direction (for example horizontal only)

Peetee06:
Thanks for reminding me of this. Will incorporate that as soon as I get the free panning to work as expected. 👍

Peetee06:
I just pushed a new update. We can now zoom and pan freely. Apart from the obvious need for refactoring, adding it to the other charts as well and testing, here are some open issue I still need to tackle.

When everything mostly works, I'll think of a good API to let users configure panning and scaling.

# Open Issues
## Axis Titles formatting is off sometimes due to double precision while scaling
Screenshot 2024-11-29 at 13 58 55

## Scaling in, panning, then scaling out leads to empty space being rendered
Screenshot 2024-11-29 at 18 10 24

## Panning on mobile doesn't precisely follow touch

https://github.com/user-attachments/assets/e01c1f64-16db-4c0d-a4c9-e7735c380e19

## Resetting scale with double tab should be animated

https://github.com/user-attachments/assets/cd860f7e-1cdb-4b70-8c44-ee762e1dd28e

## Line Chart Sample 8 throws `Bad state: Invalid`
Because the scaled values aren't mapped to one of the expected `int` values `[0, 2, 4, 6]`.

https://github.com/user-attachments/assets/8d017a11-fedb-4739-9b91-58d0dfc7d30b



## Several issues with parent scroll view
- On mobile: Can't  Y direction without first panning in X direction. Parent scroll view will scroll instead
- On web: Scrolling via Trackpad will also scroll parent scroll view. I don't have a mouse that I could test with.

Peetee06:
Just pushed the fix for **Scaling in, panning, then scaling out leads to empty space being rendered**. We're now always staying inside the user-provided min/max bounds

Peetee06:
## Zooming doesn't work on mobile when touch points are defined
It seems like the touch events are not propagated to the line chart's Gesture Detector when touch points are present on mobile devices.

Video showing me trying to zoom in on mobile:

https://github.com/user-attachments/assets/8f7c439e-5a81-4448-be37-65c3be711470

gouthamrg:
@Peetee06 @imaNNeo  I just wanna jump in here and give an example of how I have enabled zoom using a wrapper. I have been using the Zoomable_chart widget which needs to wrap the chart. I have been facing this Gesture Detector issue a lot during implemenation. Here is a simple wrapper that works in all platforms. I have enhanced a the zoomable_chart using the Listener.  I had the Gesture before but I have moved to Listener which works smoothly. I felt like I should be sharing this ( hope it helps)

- I have also implemented the feature to enable/disable y axis ( isYaxisFixed Prop)
- You can also click and drag the chart in all directions.
- Scrolling will zoom in/out in all platforms

Below is the Updated Zoomable Chart code. 


```
import 'dart:math';

import 'package:flutter/gestures.dart';

class ZoomableChart extends StatefulWidget {
  ZoomableChart({
    super.key,
    required this.maxX,
    required this.minX,
    required this.builder,
    this.isZoomPanMode = true,
    this.isYaxisFixed = false,
    required this.maxY,
    required this.minY,
    this.horizontalDragSensitivity = 0.03,
    this.verticalDragSensitivity = 0.03,
    this.zoomSensitivity = 0.0005,
  });

  final double maxX;
  final double minX;
  final double maxY;
  final double minY;
  final bool isZoomPanMode;
  final bool isYaxisFixed;
  final double horizontalDragSensitivity;
  final double verticalDragSensitivity;
  final double zoomSensitivity;
  Widget Function(double, double, double, double) builder;

  @override
  State<ZoomableChart> createState() => ZoomableChartState();
}

class ZoomableChartState extends State<ZoomableChart> {
  late double minX;
  late double maxX;
  late double minY;
  late double maxY;

  late double lastMaxXValue;
  late double lastMinXValue;
  late double lastMaxYValue;
  late double lastMinYValue;

  double _scale = 1.0;
  double _previousScale = 1.0;

  @override
  void initState() {
    super.initState();
    minX = widget.minX;
    maxX = widget.maxX;
    minY = widget.minY;
    maxY = widget.maxY;
  }

  /// Handles pointer signals for Windows platform to implement zooming functionality.
  /// Adjusts the minX, maxX, minY, and maxY values based on the scroll delta.
  handleWindowsPointerSignal(PointerSignalEvent pointerSignal) {
    if (pointerSignal is PointerScrollEvent) {
      double scrollDelta = pointerSignal.scrollDelta.dy;
      var lastMinMaxXDistance = max(maxX - minX, 0);
      var lastMinMaxYDistance = max(maxY - minY, 0);
      setState(() {
        double newMinX =
            minX - lastMinMaxXDistance * widget.zoomSensitivity * scrollDelta;
        double newMaxX =
            maxX + lastMinMaxXDistance * widget.zoomSensitivity * scrollDelta;
        if (!widget.isYaxisFixed) {
          double newMinY =
              minY - lastMinMaxYDistance * widget.zoomSensitivity * scrollDelta;
          double newMaxY =
              maxY + lastMinMaxYDistance * widget.zoomSensitivity * scrollDelta;
          if (newMinY < newMaxY) {
            minY = newMinY;
            maxY = newMaxY;
          }
        }
        if (newMinX < newMaxX) {
          minX = newMinX;
          maxX = newMaxX;
        }
      });
    }
  }

  /// Handles pointer move events for dragging.
  /// Adjusts minX, maxX, minY, and maxY based on the drag distance.
  handlePointerMove(PointerMoveEvent pointerMoveEvent) {
    if (pointerMoveEvent.buttons == kPrimaryButton) {
      var deltaDx = pointerMoveEvent.delta.dx;
      var deltaDy = pointerMoveEvent.delta.dy;
      setState(() {
        // Horizontal dragging
        double primDeltaX = deltaDx;
        if (primDeltaX != 0) {
          if (primDeltaX.isNegative) {
            minX += maxX * widget.horizontalDragSensitivity;
            maxX += maxX * widget.horizontalDragSensitivity;
          } else {
            minX -= maxX * widget.horizontalDragSensitivity;
            maxX -= maxX * widget.horizontalDragSensitivity;
          }
        }

        // Vertical dragging
        if (!widget.isYaxisFixed) {
          double primDeltaY = deltaDy;
          if (primDeltaY != 0) {
            if (primDeltaY.isNegative) {
              minY -= maxY * widget.verticalDragSensitivity;
              maxY -= maxY * widget.verticalDragSensitivity;
            } else {
              minY += maxY * widget.verticalDragSensitivity;
              maxY += maxY * widget.verticalDragSensitivity;
            }
          }
        }
      });
    }
  }

  /// Handles scale update for pinch-to-zoom functionality.
  handleScaleUpdate(ScaleUpdateDetails details) {
    setState(() {
      _previousScale = _scale;
      _scale = details.scale;

      if (_previousScale < _scale) {
        // Zoom in
        double midX = (lastMaxXValue + lastMinXValue) / 2;
        double deltaX = (lastMaxXValue - lastMinXValue) * _scale / 2;
        minX = midX - deltaX;
        maxX = midX + deltaX;

        double midY = (lastMaxYValue + lastMinYValue) / 2;
        double deltaY = (lastMaxYValue - lastMinYValue) * _scale / 2;
        minY = midY - deltaY;
        maxY = midY + deltaY;
      } else if (_previousScale > _scale) {
        // Zoom out
        double midX = (lastMaxXValue + lastMinXValue) / 2;
        double deltaX = (lastMaxXValue - lastMinXValue) * _scale / 2;
        minX = midX - deltaX;
        maxX = midX + deltaX;

        double midY = (lastMaxYValue + lastMinYValue) / 2;
        double deltaY = (lastMaxYValue - lastMinYValue) * _scale / 2;
        minY = midY - deltaY;
        maxY = midY + deltaY;
      }
    });
  }

  /// Moves the chart to center around the provided value.
  void moveChartToValue(double value) {
    print(maxX);
    var range = maxX - minX;
    setState(() {
      minX = value - range / 2;
      maxX = value + range / 2;
    });
  }

  @override
    return Listener(
      onPointerSignal: handleWindowsPointerSignal,
      onPointerMove: handlePointerMove,
      onPointerDown: (PointerDownEvent event) {
        if (event.buttons == kPrimaryButton) {
          lastMinXValue = minX;
          lastMaxXValue = maxX;
          lastMinYValue = minY;
          lastMaxYValue = maxY;
        }
      },
      behavior: HitTestBehavior.opaque,
      child: GestureDetector(
        onScaleUpdate: handleScaleUpdate,
        child: widget.builder(minX, maxX, minY, maxY),
      ),
    );
  }
}
```

Here is a simple example

```return AspectRatio(
      aspectRatio: 16 / 9,
      child: ZoomableChart(
        maxX: max(a.maxX, b.maxX),
        minX: min(a.minX, b.minX),
        minY: chartProvider.minY,
        maxY: chartProvider.maxY,
        horizontalDragSensitivity: chartProvider.horizontalDragSensitivity,
        verticalDragSensitivity: chartProvider.verticalDragSensitivity,
        zoomSensitivity: chartProvider.zoomSensitivity,
        isYaxisFixed: chartProvider.isYaxisFixed,
        builder: (minX, maxX, minY, maxY) {
          return CustomLineChart(
            minX: minX,
            maxX: maxX,
            isZoomPanMode: widget.isZoomPanMode,
            context: context,
            minY: chartProvider.isYaxisFixed ? chartProvider.minY : minY,
            maxY: chartProvider.isYaxisFixed ? chartProvider.maxY : maxY,
          );
        },
      ),
    );
  }`

**Comment by imaNNeo** [comment]
> * I have also implemented the feature to enable/disable y axis ( isYaxisFixed Prop)
> * You can also click and drag the chart in all directions.
> * Scrolling will zoom in/out in all platforms
> * You can also configure the speed of dragging (horizontalDragSensitivity, verticalDragSensitivity ) and zoom (zoomSensitivity)

Yes, I think that's related to the part of the code you mentioned.
It seems it is because of the [GestureArena](https://api.flutter.dev/flutter/gestures/GestureArenaManager-class.html) concept that we have in flutter.
So it tries to propagate the gestures to the recognizers that are registered. You can comment the part that we register `PanGestureRecognizer` to see what happens.

**Comment by Peetee06** [comment]
@gouthamrg thank you for sharing your implementation! I will use it as inspiration for the next steps 👍

@imaNNeo thanks for the suggestion. Will look deeper into the `GestureArena` and see if we can keep the current logic while also allowing the new zoom on mobile. I commented out the part with the `PanGestureRecognizer`. It allows me to pan, but weirdly, the scaling is reset when tapping a single time. Although my implementation should only do so on double tap. 

Will keep you updated with progress 👍

**Comment by gouthamrg** [comment]
Hey @Peetee06 Be aware of this problem, You might also face this while zooming, the x-axis values would overlap. 
image

Hey @imaNNeo, I’d appreciate it if you could share your suggestions here. Thanks in advance!"

**Comment by imaNNeo** [comment]
> Hey @imaNNeo, I’d appreciate it if you could share your suggestions here. Thanks in advance!"

Hey, In my opinion, that needs a separate issue that we already have #906
And it can be done in a separate PR as well.
So let's keep the scope of this PR as small as possible

**Comment by Peetee06** [comment]
The latest commit contains a refactor with a `ChartTransformationController` that has a `Matrix3` value to handle transformation and scaling. This should make the transformations more robust and the code easier to read. 

Additionally, I've added `FlScale*` events and added a `ScaleGestureRecognizer` to the `RenderBaseChart` to handle the scaling from there and work around issues where its recognizers consumed the events we want to act on. 

I had to comment out the `PanGestureRecognizer`, because all scaling gestures seemed to be considered as panning. Sadly the `Drag*` events it omits don't contain a scale value, so we can't use that. Will look into a solution for this later on.

The main issue I'm currently not sure how to tackle is the translations not behaving as one would expect. 
When being zoomed in and then panning around the chart, I would assume the chart moves exactly under my finger. This is not the current behavior. Instead, the chart lags behind.

I tried to normalize the translation based on the `LineChart` view dimensions:

```dart
// line_chart.dart line 110 - 112
final normalizedTranslationX = translationX / _viewWidth * scaledDataRangeX;
final normalizedTranslationY = translationY / _viewHeight * scaledDataRangeY;
```

But they are based on the `LineChart`'s constraints. What we need are the dimensions of the `LineChartLeaf`. Maybe you know how I can get the width and height of the `LineChartLeaf`, @imaNNeo? This way I could correctly normalize translation and make them feel natural.

Peetee06:
Update: I found some more cases that I didn't consider yet. Calculations don't work with negative values correctly, yet. Will look deeper into the maths behind this and how to properly handle transformations of min/max X/Y

imaNNeo:
>But they are based on the LineChart's constraints. What we need are the dimensions of the LineChartLeaf. Maybe you know how I can get the width and height of the LineChartLeaf, @imaNNeo? This way I could correctly normalize translation and make them feel natural.

Hi @Peetee06. Thanks for the effort you're putting in.
And sorry for my late reply. I will try to answer as quickly as possible.

LineChart doesn't know the size of the chart by itself. You (as a developer) have to provide the size (constraints), and then it can draw the content of the chart based on the provided constraints (it always chooses constraints.biggest).
For example, here we have 7 points to show as a line. And suppose the size you provide (by wrapping it with a SizedBox for example) is (700 x 300). It realizes that there should be `700 / 7 => 100`px in between the points (int x-axis). So this is the default behavior of the fl chart (and that's why you get an error if you try to put it directly in a ListView. You have to provide a size for it. For example by having an [AspectRatio](https://api.flutter.dev/flutter/widgets/AspectRatio-class.html))

More specifically, you provide the constraints and it decides how much space does it need. And it needs all of it (biggest) :)
You can read [this famous article](https://docs.flutter.dev/ui/layout/constraints) as a reference (which says: `Constraints go down. Sizes go up. Parent sets position.`)

image

And here as you see, we map the chart points (spots) into the pixels (based on the provided width and height). This way, we can draw the chart in different widths and heights:

https://github.com/imaNNeo/fl_chart/blob/269472b081baa4b8d53e734d47b6f1a35e42a428/lib/src/chart/base/axis_chart/axis_chart_painter.dart#L448-L469

So to have the chart's width and height, you can simply wrap it with a [LayoutBuilder](https://api.flutter.dev/flutter/widgets/LayoutBuilder-class.html) to get the constraints that are passing down to the chart. and as you know it always gets the biggest size, you can do this:
```dart
LayoutBuilder(
  builder: (context, constraints) {
    final chartSize = constraints.biggest;
    return LineChartLeaf(
      ...
    );
  },
);
```

I'm not sure if you should also change the chart size based on the scale events or not. (maybe you can decrease the size when it is zoomed out)

I hope this information helps you.

Peetee06:
@imaNNeo thanks so much for the explanation! I thought because of stacking the chart with the axis titles in `AxisChartScaffoldWidget` the resulting values would be off. But it **works like a charm** 👍

I don't have as much time in the coming week as I thought I have. So progress will a bit slower. I'll mainly work on improving the scaling/translating logic to get it to work correctly when data values are negative. 

Thank you again for taking the time to review and helping me with valuable insights on `fl_chart`. 🙏

Peetee06:
I have much more time in the coming week so hopefully will make good progress starting tomorrow.

The past days I tried to build the scaling and transformation logic analogous to how the `InteractiveViewer` does it. I ran into issues, because we are not working with the viewport values directly but rather the min/max X/Y values. 

I feel like I gained enough understanding now to plan out the correct calculations necessary for scaling and transforming the min/max X/Y values. 

In this [Figma file](https://www.figma.com/design/uVwnhjjCBruDW291xFJECT/fl_chart?node-id=0-1&t=SRmdhd1wXcUssfaF-1) I sketched out the possible transformation scenarios and how the adjusted min/max values relate to the original values passed to the widget. 

Next I will plan out how to properly convert viewport values like the `focalPoint` to the data space. 

I'll look for a design that clearly separates the viewport space from the data space, to increase understanding and reduce confusion and risk of errors.

Peetee06:
The newest commit contains working logic to scale charts via a Trackpad magnifying gesture. 

Previously it wasn't possible to scale around the focal point of the cursor or magnifying gesture. This commit allows proper scaling. 

# Scaling Demo

https://github.com/user-attachments/assets/ae203baa-d160-4f74-93cb-f9b7c171c90e


# Issue: Broken Animations
Because we need to update the `_lineChartDataTween` on every build to display the new min/max values, animated charts break. A possible fix I can think of is to create a wrapper widget that contains the new scaling/translation logic so as to not interfere with the tween.

https://github.com/user-attachments/assets/7c203b09-e349-40bd-a834-261f9cd5f7f7


# Next steps
1. Add translation logic for Trackpad
2. Add scaling + translation logic for Touch Gestures
3. Find a way to refactor the scaling/translation logic into a Wrapper Widget that I can test easily and makes implementation easier to maintain.

Peetee06:
# Panning via Trackpad

https://github.com/user-attachments/assets/f08e78ae-158e-41d3-8d80-3072c53d8193


# Issue: Parent ScrollView scrolls
Didn't yet find a solution for this. Will look for one later on after having finished the other tasks

imaNNeo:
Hi @Peetee06 
I'm impressed by your progress. Thanks a lot!
I checked your figma file and yes, it is exactly what we should have and it makes everything clear.

Just to mention please keep in mind that the pan/zoom feature, should be disabled by default and it is an optional feature.
(Also, we should be able to enable it only for Horizontal, Vertical or both.)

lib/src/chart/line_chart/line_chart.dart
```diff
@@ -47,17 +51,299 @@ class _LineChartState extends AnimatedWidgetBaseState<LineChart> {
...
-        data: _withTouchedIndicators(_lineChartDataTween!.evaluate(animation)),
-        targetData: _withTouchedIndicators(showingData),
-        key: widget.chartRendererKey,
+      chart: LayoutBuilder(
+        builder: (context, constraints) {
+          _chartSize = constraints.biggest;
+          return Listener(
+            onPointerSignal: _onPointerSignal,
+            child: GestureDetector(
+              onDoubleTap: resetDataAndSetState,
```
imaNNeo, line 334: Regarding this part, I think it makes more sense to handle the interaction in the renderer (just like we do for other interactions). Then, we can listen to the touch events and manage the double-tap callback.
This way, we keep the consistency

imaNNeo:
And about the animation,
Why do you need to update the tween?
What if you just transform the animated data like this?
Because you should not touch the tween, it get's updated in the [forEachTween](https://api.flutter.dev/flutter/widgets/ImplicitlyAnimatedWidgetState/forEachTween.html) callback
```dart
final transformedData = _transformData(_data);
// _lineChartDataTween = LineChartDataTween(
//   begin: transformedData,
//   end: _data,
// );
```

lib/src/chart/line_chart/line_chart.dart
Peetee06: Good point, will refactor accordingly 👍

Peetee06:
@imaNNeo thank you for the reminder about allowing to enable/disable the zooming/panning feature. I added a TODO list to the PRs description so we can keep better track of important details like these. 👍

lib/src/chart/line_chart/line_chart.dart
Peetee06: Done in 2d0c5e8

Peetee06:
I will now proceed to refactoring the code into its own widget which will hopefully resolve the current issues.

Peetee06:
The newest commits contain a refactoring that allows to more easily scale/pan bar and scatter chart. @imaNNeo what do you think about this approach? I'm not 100% happy and will look for nicer approaches. 

# Issues
## Animation is applied when scaling/panning
Because the we adjust the min/max values in the parent and then pass them to the respective chart widget, these changes are being animated. If instant scaling/panning is required, `Duration.zero` needs to be used. Allowing animations + instant scaling/panning would require to think of a completely different approach than the current one.


https://github.com/user-attachments/assets/d850596d-5a3c-4d98-8847-686d29f62ccb



## TypeError on `touchCallback`
Because the `_touchCallback` in the new `BaseInteractiveChart` uses `BaseTouchResponse` as a parameter, the `touchCallback` passed in the `BarChartSample1` throws this error
```
The following _TypeError was thrown while handling a pointer data packet:
TypeError: Instance of '(FlTouchEvent, BarTouchResponse?) => void': type '(FlTouchEvent,
BarTouchResponse?) => void' is not a subtype of type '((FlTouchEvent, BaseTouchResponse?) => void)?'
```

This is the `_touchCallback` method:
```dart
void _touchCallback(
    FlTouchEvent event,
    BaseTouchResponse? touchResponse,
  ) {
    _onBuiltInTouch(event, touchResponse);
    widget.data.touchData.touchCallback?.call(
      event,
      touchResponse,
    );
  }
```

and the `touchCallback` of `BarChartSample1`:
```dart
(FlTouchEvent event, BarTouchResponse? barTouchResponse) {
  setState(() {
    if (!event.isInterestedForInteractions ||
        barTouchResponse == null ||
        barTouchResponse.spot == null) {
      touchedIndex = -1;
      return;
    }
    touchedIndex = barTouchResponse.spot!.touchedBarGroupIndex;
  });
}
```

This error will be thrown for any `touchCallback` that doesn't use `BaseTouchResponse?` as the `touchResponse`'s type.

## Unexpected scaling behavior on bar charts
On BarCharts, x values are locked to `minX: 0` and `maxX: 1`, thus not allowing any scaling in that direction. 

I tried setting them to other values to see how the chart changes, but they seem to have no effect at all. 

Also the scaling in Y direction has unexpected behavior and the bars' background seems to ignore the clipping settings. 

I'm not yet familiar with the render logic for bar charts, so I'm not sure what's happening here.


https://github.com/user-attachments/assets/6f5da1da-3e20-4a4d-80b0-97fcd3e3b38d

Peetee06:
The newest commit allows to scale and translate bar charts vertically. 

To allow horizontal transformations I will look into how the bar chart is layed out and see how we can approach zoom/pan calculations there. 

Overall I'm not satisfied with the approach yet, and will look into nicer, simpler and more robust implementations.

Peetee06:
I tried using an `InteractiveViewer` and its transformation matrix to scale and translate the chart and titles. 

Although it looked promising and the results were almost correct, they were just slightly off. I don't completely understand the math behind it and will use the coming days to learn about it and hopefully find the correct approach.

Peetee06:
@imaNNeo I finally found an approach that I really like. 

It's still work in progress, but I can perfectly scale and pan all line charts.

I removed all other changes, so you have a clean diff to review. Please let me know what you think of this approach. 

We're now using a custom interactive viewer that differs from the original one only in that it has no `Transform` widget wrapped around the child. We don't want those transformations to be applied, because they would enlarge lines and bars, instead of adjusting the axis scale.

The current approach does the following:
1. Use the inverse of the transformation matrix to transform the chart's bounding box. This results in a bounding box larger than the chart's when zoomed in
2. Use the transformed bounding box in the renderer to calculate where to draw the chart's contents. Larger space -> content is spread out more -> zoom effect
3. Clip anything that exceeds the canvas

This way we don't have to interfere with the chart's data and get the full benefit of transformations.

The next steps will be:
1. Adjust axis titles based on the transformation. (I already have an approach stashed from previous iterations that I will try tomorrow)
2. Add parameters to allow adjusting the maxScale and for setting scaling/panning to free, horizontal or vertical
3. Write tests for everything

example/lib/presentation/samples/line/line_chart_sample1.dart
```diff
@@ -305,52 +307,54 @@ class LineChartSample1State extends State<LineChartSample1> {
...
+                color: Colors.white.withOpacity(isShowingMainData ? 1.0 : 0.5),
+              ),
+              onPressed: () {
+                setState(() {
+                  isShowingMainData = !isShowingMainData;
+                });
+              },
+            )
+          ],
+        ),
```
Peetee06, line 357: This is now wrapped in a `ClipRect`. The rest of the "changes" is formatting.

Peetee06: With 84142ba clipping is not necessary anymore, because we don't show titles that exceed the chart's range

Peetee06:
I just pushed the (hopefully) final feature commit.

Now we can scale and pan `LineChart`, `BarChart` and `ScatterChart` 🎉

I took extra care to correctly hide titles, lines and tooltips when the corresponding data exits the chart area. I checked all sample charts manually and everything looks good 🙌

In the coming days I will write tests for everything and hopefully mark this as ready for review later in the week 🤞

lib/src/chart/base/axis_chart/scale_axis.dart
```diff
@@ -0,0 +1,20 @@
+enum ScaleAxis {
```
imaNNeo, line 1: Doesn't it make sense to use [PanAxis](https://api.flutter.dev/flutter/widgets/PanAxis.html) instead?
(I see that we have `PanAxis.aligned` which is probably challenging to implement here, but we can disable the functionality for now and maybe someday we'll be able to implement that)

In this way, we're consistent with the framework.

lib/src/chart/line_chart/line_chart_painter.dart
```diff
@@ -62,12 +62,20 @@ class LineChartPainter extends AxisChartPainter<LineChartData> {
     PaintHolder<LineChartData> holder,
   ) {
     final data = holder.data;
+    if (holder.boundingBox != null) {
+      canvasWrapper
+        ..saveLayer(
+          Offset.zero & canvasWrapper.size,
+          Paint(),
```
imaNNeo, line 69: It's better to save this Paint() instance to prevent it from being re-created in every frame.

lib/src/chart/bar_chart/bar_chart_renderer.dart
```diff
@@ -9,17 +9,27 @@ import 'package:flutter/cupertino.dart';
 
 /// Low level BarChart Widget.
 class BarChartLeaf extends LeafRenderObjectWidget {
-  const BarChartLeaf({super.key, required this.data, required this.targetData});
+  const BarChartLeaf({
+    super.key,
+    required this.data,
+    required this.targetData,
+    required this.canBeScaled,
+    required this.boundingBox,
```
imaNNeo, line 17: Here we have `boundingBox` and in the axis_chart_scaffold_widget we have `chartRect`
But as I understood they're the same and they represent the virtual size that the chart is being drawn on.
So I think it makes sense to use a unique name for this concept. Because it will be confusing with the actual chart size that the user sees.

Why not using `chartVirtualSize`?

imaNNeo:
Okay, I just finished my review,
It looks very good to me. And I like the new approach. It is clean and robust.
Please take a look at those 3 reviews.

imaNNeo:
The only thing that I'm worried about, is the fact that we're calculating everything to being drawn on the part of the screen that users dont see.
Suppose that we have a single and long LineChart with 100 points, when user zooms 10x, they see only 10 points in the screen, and we have 90 points that are outside of the screen. The path is calculated using those 90 points, but being clipped in the showing part.

But in my opinion it is not blocking, we can improve it later.

lib/src/chart/base/axis_chart/scale_axis.dart
Peetee06: I agree that it would be great to align with the Framework. 

`PanAxis` only applies to panning via touch gestures. It doesn't work with scrolling/trackpad and does not prevent scaling.

For `fl_chart` we want the developer to define the axis in which they want users to be able to scale the chart. Hence `ScaleAxis` should lead to more clarity and clearly differentiate from the differing behavior of `PanAxis`.

lib/src/chart/line_chart/line_chart_painter.dart
Peetee06: good catch 👍 Added a `_clipPaint` property and used it here

lib/src/chart/bar_chart/bar_chart_renderer.dart
Peetee06: You are right, the naming is not consistent.

`chartVirtualSize` is a step in the right direction. Maybe `chartVirtualRect` would be an even better name as it's a `Rect`, i.e. Offset and Size.

What do you think?

Peetee06:
> The only thing that I'm worried about, is the fact that we're calculating everything to being drawn on the part of the screen that users dont see. Suppose that we have a single and long LineChart with 100 points, when user zooms 10x, they see only 10 points in the screen, and we have 90 points that are outside of the screen. The path is calculated using those 90 points, but being clipped in the showing part.
> 
> But in my opinion it is not blocking, we can improve it later.

You're right that this is inefficient. 
I tried the following approaches:
- Skip any point outside of the chart bounds -> led to broken paths and rendering of area below and above
- Calculate intersection point on chart edge between the current and next point -> the approaches I've tried were inaccurate, leading to paths that differed from the expected ones

I would agree with improving this later and spending some time into crafting a robust algorithm to correctly add intersection points.

lib/src/chart/base/axis_chart/scale_axis.dart
imaNNeo: Makes sense. So let's keep ScaleAxis
What do you think to add the `Fl` in front of it? (as we have this convention in the fl_chart's codebase)

lib/src/chart/line_chart/line_chart_painter.dart
imaNNeo: Nice!
Thanks

lib/src/chart/bar_chart/bar_chart_renderer.dart
imaNNeo: Yes, `chartVirtualRect` makes sense. So the name is already descriptive, but we can add a bit of explanation in the code comment.

lib/src/chart/bar_chart/bar_chart.dart
```diff
@@ -20,18 +23,55 @@ class BarChart extends ImplicitlyAnimatedWidget {
...
+        super(
           duration: swapAnimationDuration ?? duration,
           curve: swapAnimationCurve ?? curve,
         );
 
   /// Determines how the [BarChart] should be look like.
   final BarChartData data;
 
+  /// The transformation controller to control the transformation of the chart.
+  final TransformationController? transformationController;
```
imaNNeo, line 51: Does it make sense to change the name to `chartScaleController` or `scaleTransformationController`?
I mean I think we should mention the `scale` here because it will be misleading for those who don't want to use the scale feature at all.

imaNNeo: If we add the `scale` keyword, then the users can map it to the `minScale` and `maxScale` properties that we have.

imaNNeo:
I'm getting this error when I open the LineChart page (in our example app) and scroll down/up:

```
======== Exception caught by scheduler library =====================================================
The following assertion was thrown during a scheduler callback:
setState() called after dispose(): _AxisChartScaffoldWidgetState#9a704(lifecycle state: defunct, not mounted)

This error happens if you call setState() on a State object for a widget that no longer appears in the widget tree (e.g., whose parent widget no longer includes the widget in its build). This error can occur when code calls setState() from a timer or an animation callback.

The preferred solution is to cancel the timer or stop listening to the animation in the dispose() callback. Another solution is to check the "mounted" property of this object before calling setState() to ensure the object is still in the tree.
This error might indicate a memory leak if setState() is being called because another object is retaining a reference to this State object after it has been removed from the tree. To avoid memory leaks, consider breaking the reference to this object during dispose().

When the exception was thrown, this was the stack: 
#0      State.setState.<anonymous closure> (package:flutter/src/widgets/framework.dart:1168:9)
#1      State.setState (package:flutter/src/widgets/framework.dart:1203:6)
#2      _AxisChartScaffoldWidgetState._updateChartRect (package:fl_chart/src/chart/base/axis_chart/axis_chart_scaffold_widget.dart:158:7)
#3      _AxisChartScaffoldWidgetState.updateRectPostFrame.<anonymous closure> (package:fl_chart/src/chart/base/axis_chart/axis_chart_scaffold_widget.dart:144:7)
#4      SchedulerBinding._invokeFrameCallback (package:flutter/src/scheduler/binding.dart:1397:15)
#5      SchedulerBinding.handleDrawFrame (package:flutter/src/scheduler/binding.dart:1331:11)
#6      SchedulerBinding._handleDrawFrame (package:flutter/src/scheduler/binding.dart:1176:5)
#7      _invoke (dart:ui/hooks.dart:312:13)
#8      PlatformDispatcher._drawFrame (dart:ui/platform_dispatcher.dart:427:5)
#9      _drawFrame (dart:ui/hooks.dart:283:31)
====================================================================================================
```


I'm on MacOS, and this is my flutter doctor result:
```
~ » flutter doctor                                               neo@macbookpro
Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel stable, 3.27.0, on macOS 15.1.1 24B91 darwin-arm64, locale en-US)
[✓] Android toolchain - develop for Android devices (Android SDK version 34.0.0)
[✓] Xcode - develop for iOS and macOS (Xcode 16.2)
[✓] Chrome - develop for the web
[✓] Android Studio (version 2024.1)
[✓] IntelliJ IDEA Community Edition (version 2024.2.3)
[✓] VS Code (version 1.96.0)
[✓] Connected device (4 available)
[✓] Network resources

• No issues found!
```

lib/src/chart/base/axis_chart/axis_chart_scaffold_widget.dart
```diff
@@ -18,105 +29,288 @@ import 'package:flutter/material.dart';
...
+    super.initState();
+    _transformationController =
+        widget.transformationController ?? TransformationController();
+    _transformationController.addListener(_updateChartRect);
+    updateRectPostFrame();
+  }
+
+  @override
+  void dispose() {
+    _transformationController.dispose();
```
imaNNeo, line 108: I think we should not dispose the controller without checking if it is the controller comes from the user, or we created it.

As I see, in TextField for example, we have a getter that returns the user one, and if user doesn't provide, it uses the default one (just like what we do)

https://github.com/flutter/flutter/blob/4fcf6333d2170818fe63521dbe9d84f3aeaa62ab/packages/flutter/lib/src/material/text_field.dart#L971-L972

And here it only disposes the one that is created internally.

https://github.com/flutter/flutter/blob/4fcf6333d2170818fe63521dbe9d84f3aeaa62ab/packages/flutter/lib/src/material/text_field.dart#L1179

So I think the responsibility for disposing a controller, is with the creator. (because they might re-use it I think)

lib/src/chart/bar_chart/bar_chart.dart
imaNNeo: And now I see the complexity that you talked about.
It's about the Matrix4 calculation.

Is it a good idea to create a wrapper around the `TransformationController` called `FlTransformationController` that just exposes the scale property?

imaNNeo:
Okay now I've added a new chart type (Line Chart Sample 12)
It works like a charm!
I'm really happy that we have the zoom/scroll feature now. 
Thanks a lot, Peter.

https://github.com/user-attachments/assets/255aad4e-463d-4f03-8970-e615909985df

imaNNeo:
And also I have added some buttons to control the zoom/scroll programmatically.
So this is the result:

https://github.com/user-attachments/assets/afafae47-e5f3-405f-ba99-4228b2168541

There are some points that I would like to mention:
* It doesn't know about our minScale and maxScale properties. So I think we have to implement our wrapper (as I said in one of my comments) to put the scale properties there and try to clamp the user input (based on the min and max scale properties). In this way, we have a controller that has the enable/disable feature for example and user can just enable the scroll feature using that. So the controller is the single source of truth of the feature.
* As you see, when I change the properties using the controller, it is not animated. So if we have our custom controller, we will be able to add an implicit animation (later). 
Just like the implicit animation that we have in the other parts of the chart.
* And also the new custom controller, can be a place for other zoom/scale-related features.

Maybe you can have a FlScaleController, that holds an instance of TransformationController and proxies the `dispose()` method.

I'm super excited about the result btw.

imaNNeo:
I just created my branch #1807
But I don't know how can I merge it into this branch (because it is in your account).
I think you can cherry-pick those last 3 commits

lib/src/chart/bar_chart/bar_chart_renderer.dart
Peetee06: done ✅

lib/src/chart/base/axis_chart/scale_axis.dart
Peetee06: Following the convention makes sense. And it will also help prevent possible name collisions 👍 done ✅

Peetee06:
You are right that it would be nice for `fl_chart` to correct transformations so that the chart stays in-bounds. I will think about possible approaches and write an update here.

Regarding moving `minScale` and `maxScale`, as well as `scaleAxis` into a new controller that is the single source of truth:
That would also mean developers are forced to create and handle the lifecycle of the controller if they want to use the scaling/panning feature. So from a developer experience point of view, I'd say we should keep them separate to allow ease of use for simpler use cases.

imaNNeo:
> You are right that it would be nice for `fl_chart` to correct transformations so that the chart stays in-bounds. I will think about possible approaches and write an update here.
> 
> Regarding moving `minScale` and `maxScale`, as well as `scaleAxis` into a new controller that is the single source of truth: That would also mean developers are forced to create and handle the lifecycle of the controller if they want to use the scaling/panning feature. So from a developer experience point of view, I'd say we should keep them separate to allow ease of use for simpler use cases.

That's a good point.
So what if we have a wrapper around the controller, minScale, maxScale, and other properties?
For example 
```
FlScaleData
  |-> FlScaleController
  |-> minScale
  |-> maxScale
  |-> ...
```
This way, everything related to the scale feature is in a single object. But I'm not sure how can we access those properties from our FlScaleController.

Peetee06:
Wrapping a class around the transformation values would make the API cleaner and put it all in one place. I'm open to that. 

---
I've explored some options regarding ways to ensure transformations stay in bounds when the developer passes a controller to the chart. 

What comes closest to the behavior might be a ScrollController. They attach/detach to/from scroll views during which - as far as I understand -  a ScrollPosition is attached to them from the ScrollView updating the corresponding position.

In our case I could imagine something similar with the Transformation matrix being injected when the controller is attached to our chart. That way we would be able to access the current matrix from within the controller as well as possibly the chartRect which we need to correct transformation operations that would go out of bounds.

To me this sounds like an interesting approach, but I would need more time to explore it further. Also I'm pretty sure doing this right could take another 10+ hours, because of the complexity and me not being familiar with the details of ScrollPosition.

--- 
Personally I would prefer to finish this PR with the current scope and explore these improvements in future ones. This way we can get the scaling/panning into users hands and get feedback to validate our current decisions first. 

What do you think?

imaNNeo:
Thanks for your information.
I agree to merge it with the current state as we can always improve it. 
And bounding it to the chart (using controller methods) is not a blocking issue.

Could you just add the wrapper? This way we will not have any breaking changes later.

Also, the codecov issue is fixed in this branch: #1808 
Will merge it quickly

imaNNeo:
Codecov change is merged now!

Peetee06:
Sounds good! I will add the wrapper tomorrow morning. After that I will review the PR and add more tests where necessary. Then this will be ready for review 💪

Thank you for taking care of the codecov issue! 🙏

imaNNeo:
> Sounds good! I will add the wrapper tomorrow morning. After that I will review the PR and add more tests where necessary. Then this will be ready for review 💪
> 
> Thank you for taking care of the codecov issue! 🙏

Fantastic!
Hopefully, we will merge it tomorrow. 
I'm super happy about our new feature. 
Thanks for making it happen

imaNNeo:
I'm seeing it in the codecov result of this branch
image

imaNNeo:
Did you rebase `main` branch? Because I can't find the codecov commits: ce9e84905e967ccbbbbf66868ec389b3c3de91e8

Peetee06:
One thing about using a Wrapper for maxScale and minScale: If we use a wrapper LineChart and ScatterChart can't be constant anymore, because accessing the values for the assertions will only be possible at runtime then. 

Should we still use the wrapper or try to keep them const Widgets?

Peetee06:
@imaNNeo I rebased onto main. So all main branch commits should still be there. Only the commits within this branch might have changed.

imaNNeo:
> One thing about using a Wrapper for maxScale and minScale: If we use a wrapper LineChart and ScatterChart can't be constant anymore, because accessing the values for the assertions will only be possible at runtime then.
> 
> Should we still use the wrapper or try to keep them const Widgets?


I just tried to add the wrapper to see the issue, I saw that there are some assertions that can be moved inside the wrapper itself and it solved the issue.

But there was an issue with the assertion that we have in the bar chart, I had to move it inside the initState() to keep the const constructor.
And the point is that tests for the throwAssersion were broken and I couldn't fix them. (If you had time please take a look if we can bring back those tests)

You can check the changes here:
17e7b7480c585e4d780e69f3cbc5aa9ecfefb517

Feel free to cherry-pick it if you think it's good enough.
By the way, I'm not sure about the name `FlScaleData` yet. Do you have any suggestions for that?
Maybe `FlZoomData`?

imaNNeo:
> @imaNNeo I rebased onto main. So all main branch commits should still be there. Only the commits within this branch might have changed.

I'm not sure what happened, but ce9e84905e967ccbbbbf66868ec389b3c3de91e8 exists in the main, but doesn't exist in yours

Peetee06:
There was an overflow issue in the bitcoin sample that I fixed.

https://github.com/user-attachments/assets/8f958d0d-7e4e-4267-aa34-bdf457acc73e

Peetee06:
While I like grouping the transformation-related parameters in a class, I'm unsure about putting it into the chart data. To me this feels like mixing view configuration with data. The Flutter framework usually keeps these parameters with the widget, see `InteractiveViewer` which has the same parameters. In my opinion it is more intuitive to developers to make these view-related configurations on widget parameters instead of a data object and separates data from view more clearly. 

I could imagine using a `FlTransformConfig` object to wrap the scale-related parameters in. `Transform` seems like a more fitting word, because we allow scaling and translating the chart. What do you think?

To animate the scaling when using an external `TransformationController`, we need to lerp between the current and target scaling and translation values and update the matrix. Lerping the `maxScale` and `minScale` will have no effect. 

We can make the animation feature a separate PR and explore the best options. This would also be a good place to find a solution for external controllers in general.

Peetee06:
Some of the links in the new documentation won't work, because they are relative to the `main` branch. They will work after merging this into main. 👍

imaNNeo:
Everything looks good. 
I'm gonna merge it now and bump a version up!

lib/src/chart/base/axis_chart/axis_chart_scaffold_widget.dart
imaNNeo: Nice!

imaNNeo:
Hooray!
🎉

imaNNeo:
Oh I forgot to remove the long message in the squashed commit xD

Peetee06:
Niiice, super happy this is merged now 🤩 Thank you again for all the support 🙏

imaNNeo:
Thanks a lot @Peetee06 
You're one of the best contributors here and you did a great job! I really appreciate it.
I hope we can see more contributions from you in this project.

AlexanderMonneret:
amazing work, thank you @Peetee06 let me know how to send you the Donation :)
I have upgraded the lib, how can I activate the function?

Peetee06:
@AlexanderMonneret super happy it works as you expected! 🤩

if you have LinkedIn, reach out to me there. "Peter Trost", same profile picture :) Otherwise you can reach me through the email of my own little app info@emotely.de

PR #1799, opened 2024-12-05
feat: make FlSpot extend Equatable

AliAkberAakash:
Issue: 
The `FlSpot` class does not extend `Equatable`. There can be a use case where `FlSpot` object can be built from within a controller or a bloc. If it's returned as a field of state, it needs to be equatable.

example:
```
class SomeSuccessState extends Equatable {
  const SomeSuccessState(this.spotsList);

  final List<FlSpot> spotsList;

  @override
  List<Object?> get props => [spotsList];
}
```

This pr makes `FlSpot` object extend form `EquatableMixin` to address this issue. Let me know if you have any further concerns.

### DIFF
--- a/lib/src/chart/base/axis_chart/axis_chart_data.dart
+++ b/lib/src/chart/base/axis_chart/axis_chart_data.dart
@@ -450,7 +450,7 @@ class FlTitlesData with EquatableMixin {
 
 /// Represents a conceptual position in cartesian (axis based) space.
 @immutable
-class FlSpot {
+class FlSpot with EquatableMixin {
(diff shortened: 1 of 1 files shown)


imaNNeo:
Hi,
Thanks for your contribution, but the point is that for this special case we wrote the `equals ==` and `hashCode` functionalities manually.
So we don't need EquatableMixin or Equatable anymore.

Look at these lines:
https://github.com/imaNNeo/fl_chart/pull/1799/files#diff-2d7a5d668ef305a3409dd492a4dade4ab14309990fe3fb6c3acc5ef204def775R508-R520

Issue #1818, opened 2024-12-20
Allow locking zoom/pan


Peetee06:
Holding and dragging will pan the chart when it is zoomed in. This prevents touch indicators from showing while holding and dragging.

**Describe the solution you'd like**
Allow disabling scale/pan while keeping the current scaling/panning. This would allow users to zoom in, lock the chart to the current scale/pan and then hold and drag to show touch indicators.

Could look like this:

https://github.com/user-attachments/assets/74361b8d-648f-40de-9f77-cc4d59d7264a

Thanks to @AlexanderMonneret for the feedback.

Peetee06:
Thinking about this further, there are multiple things to consider:

1. As a user I want to zoom in on data and then hold and drag to display touch tooltips
In this case the transformations need to be disabled, while the current `chartVirtualRect` is kept the same.

2. As a developer I want to restrict users to transform the chart via controls only, to have more control  over how users can interact with the chart.
Therefore I want to disable the scaling/panning for users via gestures, while still allowing me to transform the chart via the `TransformationController` programatically.

Peetee06:
I found a possible solution. Please see #1819 :)

haseeb123-web:
Hi,

Thank you for adding zooming support to fl_chart! It’s a fantastic feature.

It would be great to have an option to set an initial zoom level

mustafamasri1:
https://github.com/user-attachments/assets/e44c3fd4-34f1-4d37-b957-580d62efe1d2

How i can fix this overflow issue
if i have a report for 30 days like this

Peetee06:
@haseeb123-web @mustafamasri1 you can pass your own `TransformationController` to the `FlTransformationConfig`. Initialize the controller with `TransformationController(Matrix4.identity()..scale(2.0))`. Make sure to use two dots there, because it's a cascade operation. 

Let me know if it works as you expect :) 

For anything else, look at the `AxisChartScaffoldWidget` implementation to see how we do the transforms, as well as the [Transformation documentation](https://github.com/imaNNeo/fl_chart/blob/main/repo_files/documentations/handle_transformations.md)

mustafamasri1:
Thank you, it works perfectly as expected! 🎉 .The reference to the AxisChartScaffoldWidget and the Transformation documentation was also super helpful. Thanks again!

imaNNeo:
[...]

PR #1819, opened 2024-12-20
feat: Allow locking scale/translation state

Peetee06:
Allows to disable panning and scaling.

When disabled users cannot scale/pan the chart. Transforming the chart via a `TransformationController` is still possible in this case.

Solves #1818

# Demo

https://github.com/user-attachments/assets/e1acde03-1c65-4437-b955-d0abb39bc955


imaNNeo:
That's a perfect solution.
Thanks!

PR #1822, opened 2024-12-23
Add borderRadius support to PieChartPainter

YawarOsman:
### What’s New?  
- Added `borderRadius` capability to the `PieChartPainter` for enhanced customization.  
- Ensured existing styles and functionality remain intact.

### Why?  
To provide developers with more flexibility in customizing the appearance of pie chart segments.

### Testing  
- Verified various `borderRadius` values.  
- Checked for visual regressions and ensured backward compatibility.

### DIFF
--- a/test/chart/pie_chart/pie_chart_painter_test.dart
+++ b/test/chart/pie_chart/pie_chart_painter_test.dart
@@ -614,17 +614,40 @@ void main() {
 
   group('drawSection()', () {
     test('test 1', () {
-      const viewSize = Size(200, 200);
-      final data = PieChartData(
-        centerSpaceColor: MockData.color1,
-        sectionsSpace: 10,
-        sections: [
-          PieChartSectionData(color: MockData.color1, value: 1),
-          PieChartSectionData(color: MockData.color2, value: 2),
-          PieChartSectionData(color: MockData.color3, value: 3),
-          PieChartSectionData(color: MockData.color4, value: 4),
-        ],
-      );
+       const viewSize = Size(200, 200);
+    final data = PieChartData(
+      centerSpaceColor: MockData.color1,
+      sectionsSpace: 10,
+      sections: [
+        PieChartSectionData(
+          color: MockData.color1,
+          value: 1,
+          borderSide: MockData.borderSide1,
+          borderRadius: BorderRadius.circular(10.0), // Adding border radius
+        ),
+        PieChartSectionData(
+          color: MockData.color2,
+          value: 2,
+          borderSide: MockData.borderSide2,
+          borderRadius: BorderRadius.circular(20.0), // Different border radius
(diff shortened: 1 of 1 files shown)


lib/src/chart/base/base_chart/base_chart_data.dart
```diff
@@ -17,21 +17,26 @@ abstract class BaseChartData with EquatableMixin {
   BaseChartData({
     FlBorderData? borderData,
     required this.touchData,
-  }) : borderData = borderData ?? FlBorderData();
+    BorderRadius? borderRadius,
```
imaNNeo, line 20: We don't need to have `borderRadius` property here in our base chart data. (So we don't need to have it for all the charts. And if we do, we should have it around the chart. Not something that only applies for the pie section paths)
So please remove it from here and just add it inside the `PieChartSectionData`

lib/src/chart/pie_chart/pie_chart_painter.dart
```diff
@@ -312,14 +312,46 @@ class PieChartPainter extends BaseChartPainter<PieChartData> {
...
       )
       ..style = PaintingStyle.fill;
-    canvasWrapper.drawPath(sectionPath, _sectionPaint);
+
+    // Apply border radius if specified
+    if (section.borderRadius != BorderRadius.zero) {
+      Rect bounds = sectionPath.getBounds();
+      Path roundedPath = Path()
+        ..addRRect(
+          RRect.fromRectAndCorners(
```
imaNNeo, line 338: This logic doesn't seem right to me.
And when I applied `borderRadius: const BorderRadius.all(Radius.circular(200)),` to our PieChartSample1, I saw this:
image

It only works with small values like 8, 16, ...
So it is not stable I would say

imaNNeo: I think the better idea is just to edit the shape path generation logic here:

https://github.com/imaNNeo/fl_chart/blob/01689ee2feb8ed560c80e16c19bd25f436737cb1/lib/src/chart/pie_chart/pie_chart_painter.dart#L220-L227

We might be able to use [arcTo](https://api.flutter.dev/flutter/dart-ui/Path/arcTo.html), [conicTo](https://api.flutter.dev/flutter/dart-ui/Path/conicTo.html) or [cubicTo](https://api.flutter.dev/flutter/dart-ui/Path/cubicTo.html) in [Path](https://api.flutter.dev/flutter/dart-ui/Path-class.html)

YawarOsman:
You're right, sorry for that inconsistency

imaNNeo:
No worries!
There's a new PR for this topic with a new implementation. 
So I'll close this one for now.
Thanks a lot for your contribution and idea!

Please follow it here: #2024

Issue #522, opened 2020-12-22
add labels for RangeAnnotations


guenth39:
I would like to see the possibility to pass a label to a RangeAnnotation like in [charts_flutter](https://google.github.io/charts/flutter/example/line_charts/range_annotation). And maybe an Alignment like start, center and end.

NikitaRubocki:
Did this ever get implemented? I've been digging around would also like the ability to show labels or hover capabilities to range annotations, so that on hover information about what the range is would be shown to the user.

aaronxmc:
Is there a solution to this problem ？
I need this feature as well

PR #1865, opened 2025-02-17
Add text field to range annotations

Shunt22:
We can use something like this to add text for horizontal and vertical range annotations.
Refrering to: #522.
Usage:
```dart
VerticalRangeAnnotation(
  horizontalPadding: 0,
  verticalPadding: 0,
  text: "Hello world",
  style: TextStyle(),
  rotation: 90,
  x1: 0,
  x2: 10,
  color: Colors.red,
 )

### DIFF
--- a/example/lib/presentation/samples/line/line_chart_sample8.dart
+++ b/example/lib/presentation/samples/line/line_chart_sample8.dart
@@ -129,7 +129,7 @@ class _LineChartSample8State extends State<LineChartSample8> {
               x2: 9,
               color: AppColors.contentColorBlue.withValues(alpha: 0.2),
               text: "This is a text",
-              verticalPadding: -10,
+              verticalPadding: 10,
               style: const TextStyle(
                 color: Colors.amber,
                 fontSize: 18,
@@ -141,7 +141,7 @@ class _LineChartSample8State extends State<LineChartSample8> {
             y2: 3,
             color: AppColors.contentColorGreen.withValues(alpha: 0.3),
             text: "This is a text",
-            horizontalPadding: -30,
+            horizontalPadding: 30,
           ),
         ],
       ),
--- a/lib/src/chart/base/axis_chart/axis_chart_painter.dart
+++ b/lib/src/chart/base/axis_chart/axis_chart_painter.dart
@@ -195,11 +195,11 @@ abstract class AxisChartPainter<D extends AxisChartData>
 
         drawAnnotationText(
           canvasWrapper,
-          offset: rect.centerLeft -
+          offset: rect.topCenter +
               Offset(annotation.horizontalPadding, annotation.verticalPadding),
           text: annotation.text,
           style: annotation.style,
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: CHANGELOG.md [hunk 1]
```diff
@@ -1,3 +1,6 @@
+## 0.70.3
```
imaNNeo, line 1: You don't need to touch the changelog file anymore. please revert this change

pubspec.yaml
```diff
@@ -1,6 +1,6 @@
 name: fl_chart
 description: A highly customizable Flutter chart library that supports Line Chart, Bar Chart, Pie Chart, Scatter Chart, and Radar Chart.
-version: 0.70.2
+version: 0.70.3
```
imaNNeo, line 3: You also don't need to update the version. I will do it when we have enough changes. Please read our guideline [here](https://github.com/imaNNeo/fl_chart/blob/main/CONTRIBUTING.md)

imaNNeo:
When I tried to add it in our line chart sample 8, it didn't work as I expected. So the text is somewhere else
image

imaNNeo:
Hi, thanks for the update. Functionality seems to work well! Good job.
But there are still some concerns that I'll write about in the further messages:

lib/src/chart/base/axis_chart/axis_chart_data.dart
```diff
@@ -936,6 +942,21 @@ class HorizontalRangeAnnotation with EquatableMixin {
...
+  final String? text;
+
+  /// Text style for annotation
+  final TextStyle style;
+
+  /// Text rotation in degrees
+  final double? rotation;
+
+  /// Text text padding
+  final double horizontalPadding;
```
imaNNeo: Also, padding is not what we have here. I think we can use `offset` name here. 
And I would define a single property `final Offset textOffset` for both horizontal and vertical variables (`dx` and `dy`)

lib/src/chart/base/axis_chart/axis_chart_data.dart
```diff
@@ -936,6 +942,21 @@ class HorizontalRangeAnnotation with EquatableMixin {
...
+
+  /// Vertical text padding
+  final double verticalPadding;
+
   /// Lerps a [HorizontalRangeAnnotation] based on [t] value, check [Tween.lerp].
   static HorizontalRangeAnnotation lerp(
```
imaNNeo: And please don't forget to add the new properties in the lerp() function as well.
It helps us to have transition animation for these added properties!

imaNNeo:
And my last comment:
Does it make sense to have a textAlignment  property (which is [Alignment](https://api.flutter.dev/flutter/painting/Alignment-class.html)) and defines the relative position of the text property inside the area?
I think at the moment we always draw on bottomLeft, but using this property we can make it customizable to allow people to draw it wherever they want. For example bottomCenter.

PR #1880, opened 2025-03-18
Pie Chart with rounded corners (resolves #1175)

### LINKED ISSUE #1175: Pie Chart with rounded corners
I would like to request an option to put rounded edges on a pie chart, as in the photo, in order to modernize the interface.

Screenshot 2022-09-01 at 11 51 02 AM

n-bernat:
This PR adds `borderRadius` to `PieChartSectionData` which can be used to create rounded pie charts.

The following things are still required before I can change this draft into a PR:
- [ ] Tests
- [ ] Rewrite `sectionSpace` handler (current implementation cuts-out gaps between sections, which is not feasible when we have rounded corners, it should have proper spaces). I'm not yet sure whether it will be in this PR, or a separate one, we will see how big this change will be.


imaNNeo:
Hi, thanks for contributing.
To be honest, there are a lot of changes as I see and I think there might be a simpler solution.

Also, there are some issues with the current implementation. Please take a look at the below images:

image


image

lib/src/chart/pie_chart/pie_chart_data.dart
```diff
@@ -192,6 +194,9 @@ class PieChartSectionData with EquatableMixin {
   /// Defines the radius of section.
   final double radius;
 
+  /// Defines the border radius of section.
+  final double borderRadius;
```
imaNNeo, line 198: You also forgot to add this new property in the `List<Object?> get props => []` getter

n-bernat:
Yeah, I know that there are issues, that's why it's still a draft.

What simpler solution do you see?

n-bernat:
Tbh this approach turned out to be more work than I anticipated and if there is a better way to do it I would love to start over with a simpler implementation

imaNNeo:
I think we can rewrite this path creation logic and use curve functions such as [Path.conicTo](https://api.flutter.dev/flutter/dart-ui/Path/conicTo.html), [Path.cubicTo](https://api.flutter.dev/flutter/dart-ui/Path/cubicTo.html) or [Path.quadraticBezierTo](https://api.flutter.dev/flutter/dart-ui/Path/quadraticBezierTo.html) functions

https://github.com/imaNNeo/fl_chart/blob/c427867f73f51f6582a6e8e41137a194e3ea0047/lib/src/chart/pie_chart/pie_chart_painter.dart#L220-L227

PR #1913, opened 2025-05-01
RadarChartDataにmin, maxのプロパティを追加

hiropoo:
RadarChartDataにmin, maxのプロパティを追加しました．
これにより，レーダーグラフが相対的な値の差で描画されるのではなく，min, maxに基づく絶対的な値により描画されるようにしました．


lib/src/chart/radar_chart/radar_chart_data.dart
```diff
@@ -160,12 +162,19 @@ class RadarChartData extends BaseChartData with EquatableMixin {
   /// If [isMinValueAtCenter] is true, the minimum value of the [RadarChart] will be at the center of the chart.
   final bool isMinValueAtCenter;
 
+  /// [max] and [min] are used to define the maximum and minimum values of the [RadarChart]
+  final RadarEntry? max;
```
imaNNeo, line 166: Doesn't it make sense to just have a double value for the `min` and `max` properties?
(Just like what we have for LineChart and BarChart)

lib/src/chart/radar_chart/radar_chart_data.dart
```diff
@@ -160,12 +162,19 @@ class RadarChartData extends BaseChartData with EquatableMixin {
...
+
+  /// [max] and [min] are used to define the maximum and minimum values of the [RadarChart]
+  final RadarEntry? min;
```
imaNNeo, line 169: You should add these new properties in the `props` list (in line 276)

imaNNeo:
You also need to read the [contribution guidelines](https://github.com/imaNNeo/fl_chart/blob/main/CONTRIBUTING.md). 
Because you didn't write unit tests and you didn't add documentation
Please add them

imaNNeo:
Also, I kindly ask you to write an English title and description, as it is the language we use here.

Issue #1695, opened 2024-06-22
Please, add gradient colours for the line depending on Y value for the LineChart (for example if y > 10 I need to set one color for the line, if y > 100 it should be another color)


MishaPadalka:
<img width="423" alt="image" src="https://github.com/imaNNeo/fl_chart/assets/58531288/5568523f-d444-4545-9efb-18a598f8ecab">

I mean to have something like that, when the color of the line depends on the y value of the spot

Tobbyte:
not sure what you mean. Could adding Stops to your gradient solve your problem? You could use them to define hard edges. https://css-tricks.com/books/greatest-css-tricks/hard-stop-gradients/

imaNNeo:
Yes, you can add stops to your gradient to achieve the result.
Check the [LinearGradient](https://api.flutter.dev/flutter/painting/LinearGradient-class.html) and stops property. You can use it in the `gradient` property of [LineChartBarData](https://github.com/imaNNeo/fl_chart/blob/main/repo_files/documentations/line_chart.md#linechartbardata)

Issue #285, opened 2020-04-03
Line Chart: change color gradient based on flspot.y value


vionc:
Hi, 
fist of all, this is an awesome repo, I am truly impressed by it and also by flutter.
I am really sorry if this is a duplicated issue, I have searched for a solution here but I was not fortunate enough to find what I was looking for. If it is a duplicate though, please kindly point me to the answer that I have missed. 

I am wondering whether it is possible to adjust the color gradient based on the Y values that are used as Input for the graph.
Right now, I am implementing a line chart visualizing the occupancy of a library over time. 
I would like to color the line and below line area based on how full the library is. 
(eg. library is below 50% full -> green, above 50% -> yellow, above 70% -> red).

The picture uploaded with #274  has managed to implement it for the line.
<img width="399" alt="77940020-06978980-7286-11ea-8782-adf6ba98e3b3" src="https://user-images.githubusercontent.com/63124036/78399058-7d20e800-75f4-11ea-9990-4c80102308fa.png">

I want to add the gradient below the line as well. 
This is where I am right now:
![7-F2-F090-F-37-CA-4095-8296-FD9-CA407-E264](https://user-images.githubusercontent.com/63124036/78399436-29fb6500-75f5-11ea-8b29-a5fca4d2ffa5.jpg)

Ideally, I would imagine being able to define Y-values at which the gradient changes and respective colors. 

Is this possible or would you consider making this possible?

imaNNeoFighT:
Hi,
It's my pleasure to hear you like this library,
In the [LineChartBarData](https://github.com/imaNNeoFighT/fl_chart/blob/master/repo_files/documentations/line_chart.md#linechartbardata), we have 4 fields that they are related to drawing the gradient,
`colors`, `colorStops`, `gradientFrom`, `gradientTo`.
You can set
```
gradientFrom = Offset(0.5, 0) // topCenter
gradientTo = Offset(0.5, 1) // bottomCenter
```
to have a top to down gradient, and you can play with `colorStops` to handle gradient color stops,
we are using [gradient.linear()](https://api.flutter.dev/flutter/dart-ui/Gradient/Gradient.linear.html) under the hood.
Let me know if it makes sense.

vionc:
Thank you for your quick response! 
I still don't quite understand how to do it though. Using `colorStops` I have to provide `colorStops[i]` as a List, where `i` is a number from 0.0 to 1.0 that specifies where color[i] begins in the gradient. 
[...]

PR #1925, opened 2025-05-08
feat: add `gradientArea` property for `LineChartBarData`

huanghui1998hhh:
Maybe about #285 #1695. 

**Feature**: Add `gradientArea` property to `LineChartBarData` to allow you to control the scope of gradient effects.

`LineChartGradientArea.rectAroundTheLine` (default): Gradient effect wraps precisely around the line curve only, same behavior as in the past.
`LineChartGradientArea.wholeChart`: Gradient effect applies to the entire chart area. This means you can control the relationship between your line's color and the vertical axis to a certain extent.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 1.1.0
+* **FEATURE** (by @huanghui1998hhh) Add `gradientArea` property to `LineChartBarData` to allow you to control the scope of gradient effects, #1925
+
 ## 1.0.0
 <img width="600" alt="Image" src="https://github.com/user-attachments/assets/3d8b58f4-4ce7-489f-ba45-27ece063f57c" />
 


imaNNeo:
Can you please write the unit test and update the documentation?
You can follow our guidelines [here](https://github.com/imaNNeo/fl_chart/blob/main/CONTRIBUTING.md).

huanghui1998hhh:
Sure. Anything else?

lib/src/chart/line_chart/line_chart_data.dart
```diff
@@ -324,6 +327,12 @@ class LineChartBarData with EquatableMixin {
   /// It throws an exception if you provide both [color] and [gradient]
   final Gradient? gradient;
 
+  /// Only effective if [gradient] is provided.
+  ///
+  /// It will be used to determine the area of the gradient.
+  /// If null, the gradient area will be around the line by default.
+  final LineChartGradientArea? gradientArea;
```
imaNNeo, line 334: Can you please make it non-nullable and set the default value in the constructor itself?
(Just like most places in our code base to keep the consistency)

This way, we can always find the default value in the constructors.

imaNNeo:
> Sure. Anything else?

Everything looks good to me.
Please fix my only comment, then we're good to go.

lib/src/chart/line_chart/line_chart_data.dart
huanghui1998hhh: OK~

imaNNeo:
Perfect!
Is there any issue that it solves?
Just to mention it in the changelog and PR's description.

huanghui1998hhh:
OK! Let's go.

imaNNeo:
Perfect!
Thanks a lot

PR #1932, opened 2025-05-14
Update color_extension.dart

SoumenDas01:


imaNNeo:
`color.alpha` is deprecated. Why do you want to use it?

PR #1933, opened 2025-05-15
feat: Add Gauge Chart

glemartret:
Updated code from https://github.com/imaNNeo/fl_chart/pull/1424 (credits to @FlorianArnould)

The implementation of the gauge chart (discussed here https://github.com/imaNNeo/fl_chart/issues/199)


imaNNeo:
Can you please rebase the branch?

glemartret:
It should be OK, I also removed files not concerned by this PR

imaNNeo:
Okay, now I found a little bit of time to take a look at this one. Here I have some concerns, please have a look:

I tried to pass only one color to the `VariableGaugeColor`, it crashed, and after more investigation, I realized we also have a `SimpleGaugeColor`
So I think we can just have a single class (no abstraction) for this case and if the colors list has only one color, it's a simple drawing color, but if it has multiple colors, it is a variable color. So we can calculate the `limits` under the hood by dividing the whole space by the number of colors. Or user can provide a custom `limits` property with the customized numbers. In this way, we should throw an exception if the limits length is not equals to the colors length. (Just like other places that we have gradients and stops)

```dart
valueColor: VariableGaugeColor(
  colors: [
    AppColors.contentColorYellow,
  ],
  limits: [],
),
```
Or if you want to keep the abstraction, please let me know what other types of `GaugeColor` we can have?
(In this case, I still think that our `VariableGaugeColor` should be able to work with a single color)

imaNNeo:
As I see, when we have a `VariableGaugeColor`, it draws the stop points as a circular indicator.
How can I disable/customize them?
Because in the beginning, I found them odd here.

image

imaNNeo:
I also have a suggestion for the `startAngle` and `endAngle` properties (because they are difficult to set)
I suggest having something like `startDegreeOffset` (just like what we have in the PieChart that is 0 by default -> start from right center), then we can have `direction` that can be clockwise and counter-clockwise (I'm not sure if we already have a data structure for this one or not). Then we can have a `sweepAngle` or `sweepDegrees` that determines the lengths of the arc. I think it's more similar to how our PieChart works (because of `startDegreeOffset`) and I think it is easier to work (from the developer's perspective).

Please let me know your thoughts

rsnider19:
any updates on this? I'd love to move off of syncfusion

glemartret:
@imaNNeo I’ve rebased and updated the PR following your feedback.

example/lib/presentation/samples/gauge/gauge_chart_sample3.dart
```diff
@@ -0,0 +1,61 @@
...
+      padding: const EdgeInsets.all(28),
+      child: Column(
+        children: [
+          SizedBox(
+            width: 250,
+            height: 250,
+            child: GaugeChart(
+              GaugeChartData(
+                value: _value,
+                valueColor: GaugeColor(
```
imaNNeo: I don't understand why we need this GaugeColor() object for setting a valueColor?
I'm thinking about having a simple `Color valueColor` parameter that can be different colors based on the different values that user sets.
For example:

```
double _value = // a number that can change runtime

GaugeChartData(
  value: _value,
  valueColor: switch (_value) {
    <= 0.5 => Colors.green,
    <= 0.8 => Colors.yellow,
    _ => Colors.red,
  },
)
```

Then we can get rid of the `GaugeColor` completely to have a simpler/maintainable logic

lib/src/chart/gauge_chart/gauge_chart_painter.dart
```diff
@@ -0,0 +1,271 @@
...
+    _backgroundPaint = Paint()..isAntiAlias = true;
+    _valuePaint = Paint()..isAntiAlias = true;
+    _tickPaint = Paint();
+  }
+
+  late Paint _backgroundPaint;
+  late Paint _valuePaint;
+  late Paint _tickPaint;
+
+  _GaugePosition? _gaugePosition;
```
imaNNeo: It's not a good practice to cache objects in a painter.
I see your point, you didn't want to recalculate it in the handleTouch method. I know we're already doing something similar in BarChartPainter, but there we have a more complicated logic (iterating over all groups and bars). 
But here, the performance gain is negligible, and it's okay to just re-calculate it.

example/lib/presentation/samples/gauge/gauge_chart_sample3.dart
imaNNeo: This way, the user is also able to change the color independent of the value itself.

lib/src/chart/gauge_chart/gauge_chart_data.dart
```diff
@@ -0,0 +1,319 @@
...
+  inner,
+  outer,
+  center,
+}
+
+@immutable
+class GaugeTicks with EquatableMixin {
+  const GaugeTicks({
+    this.count = 3,
+    this.radius = 3.0,
```
imaNNeo: The way we implemented `GaugeTicks` is not customizable/scalable.
I suggest to have something like:
```dart
GaugeTicks(
  double count,
  GaugeTickPosition position,
  ...
  GaugeTickPainter painter,
)
...

abstract class GaugeTickPainter {

  // abstract draw method

}


GaugeTickCirclePainter extends GaugeTickPainter {
  double radius
  Color color
   ... // overrides the draw method and renders a circle
}
```

Please take a look at `FlDotPainter()` to get the idea.

lib/src/chart/gauge_chart/gauge_chart_data.dart
```diff
@@ -0,0 +1,319 @@
...
+    required this.color,
+    this.position = GaugeTickPosition.outer,
+    this.margin = 3,
+    this.showChangingColorTicks = true,
```
imaNNeo: I don't understand what `showChangingColorTicks` does.
Maybe it is related to the color topic that I already talked about. We can move this kind of logic outside of the library.

lib/src/chart/gauge_chart/gauge_chart_data.dart
```diff
@@ -0,0 +1,319 @@
...
+
+/// [GaugeChart] needs this class to render itself.
+///
+/// It holds data needed to draw a gauge chart
+class GaugeChartData extends BaseChartData with EquatableMixin {
+  GaugeChartData({
+    this.strokeCap = StrokeCap.butt,
+    this.backgroundColor,
+    required this.valueColor,
+    required this.value,
```
imaNNeo: The current implementation has the limitation of having only one gauge line/color. But I guess in the future we'll get the request to have multiple of them. 
So to prevent a breaking change, I suggest changing the structure to support having a list of them.

Something like this:

Image

This is my idea:

```
GaugeChartData(
    minValue: 0, // default is 0.0
    maxValue: 1000, // default is 1.0, but it lets users have scaled values based on their value range
    startDegreeOffset: -45, // we can keep it as a default value, I think!
    sweepAngle: 270, // we can keep it as a default value, I think!
    sectionsSpace: 4,  // gap between auto-stacked sections
    sections: [
      GaugeSection(
        value: 370, // it is 37% of the arc (based on min, max values)
        valueColor: Colors.red,
        backgroundColor: Colors.grey.shade200, // it is nullable, users can have transparent arc, or a grey background
        width: 25, // this is the thickness of the section,
        strokeCap,
      ),
      GaugeSection(
        value: 700, // it is 70% of the arc (based on min, max values)
        valueColor: Colors.amber,
        width: 20,
        strokeCap,
      ),
    ],
    // other top-level properties.
  )
```

imaNNeo (changes requested):

imaNNeo:
Hi there, I just wrote some comments. 
Please have a look.

As it is a new chart type, I would like to put more time/effort into it to ensure it is future-compatible and that we won't have breaking changes. That's why I'm a bit strict about it. 
And just to mention that I have some free time these days. If you think you don't have enough time to put this effort, I can take over and continue your branch. (You already did a lot of things).

Just let me know what works better for you.

glemartret:
Hey, no problem, I won't be able to work on it at least today, maybe tomorrow or in the week end. 
Feel free to update it if you want, I'm not the original author of this new chart type, I've just updated it to work with new version of your package as the original PR wasn't

imaNNeo:
Alright, so I would like to continue your work and do the final improvement.
Because I have a limited free time now, so I can fully focus on it.

imaNNeo:
Alright, we're almost done, here are 4 samples that we have created:

| Sample1 | Sample2 | 
|--------|--------|
|<video src="https://github.com/user-attachments/assets/41460876-c8d9-44ec-bed8-5bdcde3d52c4" controls preload></video> | <video src="https://github.com/user-attachments/assets/7ee90e13-2916-4707-a24b-ed7e4e25dc7f" controls preload></video> |


| Sample3 | Sample4 | 
|--------|--------|
|<video src="https://github.com/user-attachments/assets/3a645190-e626-48dc-99e3-f3dd335a2eb2" controls preload ></video> | <video src="https://github.com/user-attachments/assets/c5884c33-cfbb-4ae9-bacc-55386288a813" controls preload></video> |

imaNNeo (approved):

imaNNeo:
Thanks @glemartret  and @FlorianArnould

PR #1949, opened 2025-06-26
Creating eventType prop to prevent using .runtime type or similar

W3ND31:


test/chart/base/render_base_chart_test.dart
```diff
@@ -139,6 +139,7 @@ void main() {
         chart.handleEvent(pointerHoverEvent, hitTestEntry);
 
         expect(testEvent, isA<FlPointerHoverEvent>());
+        expect(testEvent.eventType, equals('FlPointerHoverEvent'));
```
imaNNeo, line 142: What's the problem with this one instead?:
```dart
expect(testEvent, isA<FlPointerHoverEvent>());
```

PR #1953, opened 2025-07-09
Add shrink wrap for bar chart when overflow

### LINKED ISSUE #1952: How to get rid of the extra spacing at the chart's left/ride side?
Hi,

I'm using the example you provided, bar_chart_sample1. When putting the chart in a `SizedBox(width: 300, height: 300)` and making the chart's background color red, there is spacing at the left and right side (see the yellow markers):

Image

I have not found anything to get of that, so the chart just occupies the space it requires.

Is there any property I've overlooked?

Thanks!

alettsy:
This pull request is to resolve #1952.

This should allow the user to set a property to make the first and last bars touch the sides of the container when they would otherwise overflow and get auto-spaced (in `spaceEvenly()`).

When `shrinkWrapOnWidthOverflow` is set to `true`, the first and last bars in the chart should be touching the sides instead of having space at the start/end. 

The bars should still be spaced evenly.

## Checklist
- [X] Unit tests pass
- [X] Manually tested
- [X] Formatted correctly
- [X] Doc comments

## Screenshots:

With `shrinkWrapOnWidthOverflow = true`:
image

With `shrinkWrapOnWidthOverflow = false`:
image


alettsy:
I have pushed changes to address the Copilot code review 😊

lib/src/chart/bar_chart/bar_chart_data.dart
```diff
@@ -73,6 +74,10 @@ class BarChartData extends AxisChartData with EquatableMixin {
   /// Apply space between the [barGroups].
   final double groupsSpace;
 
+  /// If width of chart would overflow the container, then space evenly without
+  /// adding space at the start and end of the chart.
+  final bool shrinkWrapOnWidthOverflow;
```
imaNNeo, line 79: You should add this property inside the `props` as well (line 169)

imaNNeo:
Does it matter what type of Alignment we're using for `BarChartAlignment`?
Or it just works in all the cases?

alettsy:
Hi 👋

I have added the property into `props`, as requested.

As for your question:

The `shrinkWrapOnWidthOverflow: true` only applies when the items would overflow the chart area, which causes the `spaceEvenly` function to get called.

This case happens when `BarChartAlignment` is `start`, `end`, or `center` and the total width of all items and spacing is greater than the chart width (`tempX > viewWidth`), or when it's set to `spaceEvenly`. When `BarChartAlignment` is `spaceAround` or `spaceBetween`, the spacing is calculated differently than the `spaceEvenly` function.

This can be seen in `lib/src/extensions/bar_chart_data_extension.dart`.

Please let me know if there are any other issues 😊

alettsy:
I added a test that should make for 100% coverage now 😊

imaNNeo:
To be honest, I don't see this flag as the fix to the bug that happens when you define `start`, `end` or `center` with a spacing that leads to overflow.
I think when it happens, we should render them edge-to-edge (just like when `shrinkWrapOnWidthOverflow` is true in your implementation). So we simply say if the required width is more than the available width, we can set spaceBetween (under the hood). 
Or maybe we can define an `overflowFallbackAlignment` that is `spaceEvenly()` by default and users can change it to `spaceBetween()`.

What do you think?

alettsy:
Yeah, that makes sense if it just automatically applies if there would be overflow, rather than setting the property manually. 

My thought was that some people might not want the bars to touch the edge so having a property to alter the behaviour when there is an overflow accounted for that. 

But for ease-of-use, I think you're right it's better to let it get worked out behind the scenes when there is an overflow, and users can calculate the appropriate spacing based on their container size if they want to avoid the overflow behaviour. 

I will update it 😊

alettsy:
Is that what you had in mind? 😊

alettsy:
Hi, @imaNNeo 👋 I didn't get a response from you in a few months, so I was wondering if this PR is complete or needs to be updated again. Please let me know! Thank you 😊

imaNNeo:
Hey, sorry for my delayed reply.
I see that `spaceEvenly` functionality is broken now.
Can you please fix it and also give me a reproducible code for the main issue that you initially had?
Then I can check the functionalities!

| | ❌ spaceEvenly  | ✅ spaceBetween  |  ✅ spaceAround | 
|---|---|---|---|
|  before  |  image | image  |  image | 
| after  | image |  image | image  |

imaNNeo:
Closing due to inactivity. Feel free to re-open it (or create a new PR)

Issue #1954, opened 2025-07-09
Feature Request: Add gradient support to FlDotPainter for custom dot rendering


mohhamad-esmaili:
Hi Dear Iman,

First of all, thank you for the great work on `fl_chart` — it's one of the best charting solutions for Flutter and we've really enjoyed using it.

While customizing line chart dots using `FlDotData` and `getDotPainter`, I noticed that there's currently no built-in support for applying gradient styles to the dots. Right now, we can only use solid colors in `FlDotCirclePainter` or other `FlDotPainters`, which limits the visual styling options — especially when trying to highlight important data points (like the last point) with gradients.

## Feature Request:

I’ve already implemented a custom `GradientDotPainter` in my project using a custom painter, and it works well. I’d be happy to contribute this as a Pull Request if you're open to the idea.

Adding this feature would unlock more flexibility and polish in custom charts and improve visual storytelling.

Thanks again for your awesome work and for considering this request.

Best regards.

<img width="289" height="198" alt="Image" src="https://github.com/user-attachments/assets/a3e88fbc-6892-4712-9d51-cc5ffb7f80ef" />

PR #1960, opened 2025-07-20
Add GradientFlDotPainter for square and circle line chart

mohhamad-esmaili:
This pull request introduces two custom painters for the FlChart library:

## FlDotGradientCirclePainter:
This class extends FlDotPainter and adds support for drawing circles with gradient fills and optional stroke. It accepts parameters like:

- `radius`: Customizes the radius of the circle.
- `gradient`: Specifies the gradient fill for the circle.
- `strokeWidth`: Adds a stroke around the circle.
- `strokeColor`: Customizes the stroke color.

The circle is drawn with a gradient fill, and a stroke is added if the `strokeWidth` is greater than zero. The painter also supports hit testing and interpolation between two painters.

## FlDotSquarePainter:
Similar to `FlDotGradientCirclePainter`, this painter allows drawing squares with solid colors and optional strokes. It accepts:

- `color`: The fill color for the square.
- `size`: Customizes the size of the square.
- `strokeWidth`: Adds a stroke around the square.
- `strokeColor`: Customizes the stroke color.

The square can be rendered with a stroke and is drawn with a solid color. Like the circle painter, it also supports hit testing and interpolation.

These two painters can be used for custom dot rendering with `FlDotPainter`, allowing more flexible styling options for users of the chart library. I already opened an issue about this, check #1954

### DIFF
--- a/lib/src/chart/base/axis_chart/axis_chart_data.dart
+++ b/lib/src/chart/base/axis_chart/axis_chart_data.dart
@@ -1553,6 +1553,87 @@ class FlDotCirclePainter extends FlDotPainter {
 }
 
 /// This class is an implementation of a [FlDotPainter] that draws
+/// a gradient circled shape
+class FlDotGradientCirclePainter extends FlDotPainter {
+  /// The gradient of the circle is determined determined by [Gradient],
+  /// [radius] determines the radius of the circle.
+  /// You can have a stroke line around the circle,
+  /// by setting the thickness with [strokeWidth],
+  /// and you can change the color of of the stroke with [strokeColor].
+  const FlDotGradientCirclePainter({
+    required this.radius,
+    required this.gradient,
+    this.strokeWidth = 0,
+    this.strokeColor = Colors.white,
+  });
+
+  /// The fill Gradient to use for the circle
+  final Gradient gradient;
+
+  /// Customizes the radius of the circle
+  final double radius;
+
+  /// The stroke color to use for the circle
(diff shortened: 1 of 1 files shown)


imaNNeo:
Hi, thanks for your PR
I think we can use our existing `FlDotCirclePainter` and `FlDotSquarePainter` to reduce the complexity (have fewer classes)
At the moment they just accept a `color` property, but you can add a `gradient` property beside the `color`. So it works exactly like how we handled it in other places, such as:
* [LineChartBarData](https://github.com/imaNNeo/fl_chart/blob/main/repo_files/documentations/line_chart.md#linechartbardata)
* [BarAreaData](https://github.com/imaNNeo/fl_chart/blob/main/repo_files/documentations/line_chart.md#barareadata)

So, if you provide the color, color applies and if you provide the gradient, gradient applies. If you provide both, it throws an error.

And we have this helper method to configure the paint class using color or gradient:
https://github.com/imaNNeo/fl_chart/blob/3a3484c751812a8f9083c9a14f6cd6eb28c36271/lib/src/chart/line_chart/line_chart_painter.dart#L1049-L1066

mohhamad-esmaili:
Yes i see, but i needed gradient for the dot and points so i added this, if we are not looking more codes like this so you're right and close PR

imaNNeo:
Yes, I agree that having gradients in our dots and points is a good idea.
I just gave my suggestion to reuse the existing classes.
If you're not able to work on it, we can close it then.

PR #1979, opened 2025-08-26
feat: add gradient support for BarChartRodStackItem

### LINKED ISSUE #919: Include gradient field to BarChartRodStackItem
I can see I can give a gradient color to bar chart. But, if there are multiple stacked bar charts, there is no way to give a gradient for each **BarChartRodStackItem**.  It would be nice to have this feature!

kamilJ96:
Resolves #919 

This PR adds a `gradient` property on `BarChartRodStackItem`, which displays a stacked chart rod with a gradient if provided.

Saw there was an attempt at this already, but no activity on that for a few weeks now and I'm needing this feature and impatient 😛 

Unfortunately [dart doesn't support both optional positional arguments and named arguments](https://github.com/dart-lang/language/issues/1076), so the constructor signature is not ideal, however this avoids a breaking change as opposed to updating `color` to a named argument. If there's a better way, or we make that breaking change, let me know.

I've also updated a sample bar chart to show the effect.

Screenshot 2025-08-26 at 11 09 26 pm

### DIFF
--- a/example/lib/presentation/samples/bar/bar_chart_sample5.dart
+++ b/example/lib/presentation/samples/bar/bar_chart_sample5.dart
@@ -169,7 +169,16 @@ class BarChartSample5State extends State<BarChartSample5> {
             BarChartRodStackItem(
               0,
               value1,
-              AppColors.contentColorGreen,
+              null,
+              gradient: LinearGradient(colors: [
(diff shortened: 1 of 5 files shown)


imaNNeo:
Simple and neat!
Thank you very much for your nice PR

PR #1980, opened 2025-08-29
fix: prevent curve overshooting

bcheidemann:
Fixes https://github.com/imaNNeo/fl_chart/issues/1722

Based on https://github.com/imaNNeo/fl_chart/issues/1722#issuecomment-2267489789

**Before (with `preventCurveOvershooting: true`)**

image

**After (with `preventCurveOvershooting: true`)**

image

**Before (with `preventCurveOvershooting: true` and `preventCurveOvershootingThreshold: 0`)**

image

**After (with `preventCurveOvershooting: true` and `preventCurveOvershootingThreshold: 0`)**

image


imaNNeo:
We have a new implementation of a curved line,
Can you please check if that already solves your issue or not?
It's not merged yet, but you can take a look at it here:
#2000

PR #1989, opened 2025-09-05
feat: Recognize secondary button pointer events

enwi:
This pull request adds support for distinguishing between primary and secondary button interactions in chart touch events. The main changes involve creating special `secondary` event classes and wiring up the appropriate gesture recognizer callbacks to handle secondary button events.

**Touch event model enhancements:**

* Added `secondary` touch event classes (`FlTapDownSecondaryEvent`, `FlTapCancelSecondaryEvent`, `FlTapUpSecondaryEvent`, `FlLongPressSecondaryStart`, `FlLongPressMoveSecondaryUpdate`, `FlLongPressSecondaryEnd`) in `fl_touch_event.dart` to indicate if the event was triggered by a secondary button.

**Gesture recognizer integration:**

* Updated `render_base_chart.dart` to handle secondary tap and long press gestures by wiring up the corresponding gesture recognizer callbacks.

These changes enable the chart to differentiate between primary and secondary pointer interactions, which is useful for supporting right-click or other alternate input methods.


imaNNeo:
Hi Owen,
Thanks for your contribution.

I think it's better to define different event types for the secondary events (instead of adding a flag inside the current events).
And the reason is the fact that it affects the current logic that users have. Suppose that they implemented a logic that only accepts the primary tap. After updating to the new version, they also support the secondary tap. (I know they can update their logic to check if it is a primary tap or not, but that's not backward compatible)

So I just suggest defining new event types for secondary events, then anyone can support the secondary button, and the current implementation remains untouched.

Thanks!

enwi:
Hey @imaNNeo I have updated the PR as you suggested 🙂

imaNNeo:
Perfect! Thanks for implementing the new approach.
It looks good to me, but before merging, there are a couple of things to mention:
* To keep the consistency with the tap callbacks in Flutter itself, I suggest using the `Secondary` keyword as a prefix, instead of a postfix. For example: 
  * FlTapDown**Secondary**Event -> Fl**Secondary**TapDownEvent
  * FlTapCancel**Secondary**Event -> Fl**Secondary**TapCancelEvent
  * FlLongPress**Secondary**Start -> Fl**Secondary**LongPressStart
  * FlLongPressMove**Secondary**Update -> Fl**Secondary**LongPressMoveUpdate
  * ...
* Can you please update the documentation in base_chart.md?
* Is there any possibility of writing a unit test for it? (I know it is deep and low-level. But there might be a way to test that

enwi:
@imaNNeo I have updated the code as you suggested. As there do not exist any tests for other events currently I am not sure if I find the time to write them.

imaNNeo:
Thanks for your contribution!

Issue #1223, opened 2023-01-27
Time series animation issue


hamishjohnson:
https://user-images.githubusercontent.com/9589803/215057034-7cefb8bf-1dd2-449e-a4c3-6007be0ea43a.mov

Time series charts cause strange looking graphs when changing the timerange during animation. Is there a way to configure the animation so that it works differently? Or is there no way to change this behavior?

imaNNeo:
That sounds weird :))
What about just removing that annoying line? I think it makes it way better.

PhilippS93:
Is there a workaround. I have currently disabled animation because it looks weird, when the date range changes.

orkun1675:
I'm facing the same problem. We have three graphs: daily, weekly, monthly. Changing between them causes this weird behaviour as displayed above.

PR #1994, opened 2025-09-10
Line chart appear animation

Michal-MK:
Possibly closing #1223 as we also had this issue in our project so I implemented this and now ported it to newest fl_chart to see if it can be merged.

https://github.com/user-attachments/assets/bfad8f49-1668-464d-9b5e-933523fb911e

So far no tests and no coverage stuff (I read the contributors guide :D)

If this is accepted as a solution I will look at that and address it in the PR.


imaNNeo:
Hi, thanks for your PR and for implementing the loading animation.
Sorry for my late reply.

To be honest, I don't think this solution is scalable and customizable.
I would like to implement it in a way that we can scale it, and users are able to customize it.
These are the comments I got from my LLM:

> 1. LineChartPaintHolder and LineChartEntryAnimation are line-chart-specific, but they're
>   used in base classes like AxisChartPainter and AxisChartScaffoldWidget that are shared
>   across all chart types. Let's keep the base layer chart-type-agnostic — you can move these
>   to the line chart package and use polymorphism (e.g. override a getter) instead of the if
>   (holder is LineChartPaintHolder) type check.
> 2. The clipping logic is repeated 4 times in drawBarLine, drawDots, drawBelowBarPath, and
>   drawAboveBarPath. Let's extract it into a helper.
> 3. dependency_overrides for vector_math shouldn't be in a published library's pubspec.yaml —
>    it can cause issues for downstream users.
> 4. The naming is inconsistent — the same concept is called LineChartEntryAnimation,
>   animationType, appearAnimationType, and entryAnimation in different places. Let's pick one
>   name.
> 5. No tests for the actual animation behavior (clipping at different progress values,
>   targetData usage during slide, dots hidden in leftToRightSlideWithoutPoints). The current
>   test changes are just mechanical PaintHolder → LineChartPaintHolder updates.
> 6. The enum values bake "leftToRight" into the name, which would make adding other
>   directions awkward later. A config object separating direction from content options might
>   work better — but just a thought, happy to discuss.

And now there are a lot of conflicts. So I'll close it for now.

Thanks again!

Michal-MK:
Oh hey this PR existed :D No worries, maybe you can take something from it and implement it better.

imaNNeo:
Haha!
Sorry, sometimes I have to do this.
But I'll keep this feature in my mind ot implement asap. We should find a simpler and more scalable solution.

Thanks for your contribution btw, it definitely gives some idea to me (and my LLM :)

PR #1996, opened 2025-09-12
feat: candlestick chart support trade indicator

artsooter:
about https://github.com/imaNNeo/fl_chart/issues/1995
---
 I have added a new `tradeSignalType` property to `CandlestickSpot`, which includes four types:
  buy/sell/trade/none. 
  
  When the type is not `none`, it will display a corresponding trading
  signal above the candlestick based on the different types

such as
image

### DIFF
--- a/example/lib/presentation/samples/candlestick/candlestick_chart_sample1.dart
+++ b/example/lib/presentation/samples/candlestick/candlestick_chart_sample1.dart
@@ -151,12 +151,24 @@ class CandlestickChartSample1State extends State<CandlestickChartSample1> {
                           .map((entry) {
                         final index = entry.key;
                         final data = entry.value;
+
+                        // Add different signal types for testing
+                        TradeSignalType signalType = TradeSignalType.none;
+                        if (index % 8 == 0 + (_currentMonthIndex % 8)) {
+                          signalType = TradeSignalType.buy;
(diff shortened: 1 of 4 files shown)


imaNNeo:
This approach is not scalable, and it just applies to the signal type.
For example, if someone wants to have a similar feature for just indicating important points, we have to develop a new feature.
So I would like to suggest implementing it in a generic way,
For example, you can define a concept called candleIndicatorPointer, and it is something that we draw on top (or bottom) of the candles (using an enum, for example). Then we can have an abstract painter inside that, and we can define a CandleBalloonPainter (which is what you already want.
So it has a text resolver that you can customize the text that it shows, and you can implement your usecase to show buy, sell, ... indicators. 

Someone else can use it to show different wordings for a different concept.
This way, the balloon painter is reusable for any other use cases. We can also define new painters for different shape of indicators.

I know it makes it more challenging to implement. But I'm sure it aligns with our goal: to be fully customizable.
I appreciate your understanding.
Thanks!

artshooter:
Hello, I sincerely apologize for the inconvenience. The author account of this PR has been disabled and is no
   longer available (@artsooter is my old account).

  I've submitted a new PR incorporating your feedback. Thank you!
  
  new PR (https://github.com/imaNNeo/fl_chart/pull/2032)

imaNNeo:
Thanks for letting me know. It would be nice if you add the image and the full description as well

PR #2004, opened 2025-09-22
feat: Add stacked segment support for PieChart sections

jay-k98:
As described in https://github.com/imaNNeo/fl_chart/issues/2003 a "Stacked Pie Chart" was requested. This Pull Request aims to add this feature.

# Solution
The Implementation of [StackedPieChart] is very similar to [PieChart]. The biggest difference is drawing a Section with multiple Segments.

The radius of a segment is calculated like this:

SegmentRadius = SectionRadius * SegmentValue / TotalSectionValues

The sweep angles are then calculated like in [PieChart] 

# Todo's and Improvements
- Add drawing strokes
- Extend the Demo (add icon and more examples)
- Add tests
- Extract duplicate logic from [PieChart] and [StackedPieChart]

### DIFF
--- a/lib/src/chart/pie_chart/pie_chart_painter.dart
+++ b/lib/src/chart/pie_chart/pie_chart_painter.dart
@@ -22,8 +22,6 @@ class PieChartPainter extends BaseChartPainter<PieChartData> {
   PieChartPainter() : super() {
     _sectionPaint = Paint()..style = PaintingStyle.stroke;
 
-    _sectionSaveLayerPaint = Paint();
-
     _sectionStrokePaint = Paint()..style = PaintingStyle.stroke;
 
     _centerSpacePaint = Paint()..style = PaintingStyle.fill;
@@ -33,7 +31,6 @@ class PieChartPainter extends BaseChartPainter<PieChartData> {
   static const double _kRadiusSafetyMargin = 0.499;
 
   late Paint _sectionPaint;
-  late Paint _sectionSaveLayerPaint;
   late Paint _sectionStrokePaint;
   late Paint _centerSpacePaint;
   late Paint _clipPaint;
@@ -114,34 +111,15 @@ class PieChartPainter extends BaseChartPainter<PieChartData> {
       final sectionDegree = sectionsAngle[i];
 
       if (sectionDegree == 360) {
-        final radius = centerRadius + section.radius / 2;
-        final rect = Rect.fromCircle(center: center, radius: radius);
-        _sectionPaint
-          ..setColorOrGradient(
-            section.color,
-            section.gradient,
-            rect,
-          )
-          ..strokeWidth = section.radius
-          ..style = PaintingStyle.fill;
-
-        final bounds = Rect.fromCircle(
-          center: center,
-          radius: centerRadius + section.radius,
+        drawSegments(
+          canvasWrapper,
+          section,
+          sectionDegree,
+          centerRadius,
+          0,
+          center,
         );
-        canvasWrapper
-          ..saveLayer(bounds, _sectionSaveLayerPaint)
-          ..drawCircle(
-            center,
-            centerRadius + section.radius,
-            _sectionPaint..blendMode = BlendMode.srcOver,
-          )
-          ..drawCircle(
-            center,
-            centerRadius,
-            _sectionPaint..blendMode = BlendMode.srcOut,
-          )
-          ..restore();
+
         _sectionPaint.blendMode = BlendMode.srcOver;
         if (section.borderSide.width != 0.0 &&
             section.borderSide.color.a != 0.0) {
@@ -152,7 +130,9 @@ class PieChartPainter extends BaseChartPainter<PieChartData> {
           canvasWrapper
             ..drawCircle(
               center,
-              centerRadius + section.radius - (section.borderSide.width / 2),
+              centerRadius +
+                  section.totalRadius -
+                  (section.borderSide.width / 2),
               _sectionStrokePaint,
             )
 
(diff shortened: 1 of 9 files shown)


imaNNeo:
Hi, thanks for your contribution.
I think we can use the current implementation of the PieChart to just add the stacked logic into it.
This way, we reuse a lot of code that we already wrote.

Just like what I did for the stacked bar chart here: #112

Please let me know your thoughts on it.

jay-k98:
Hi, yeah thank you for responding! I wanted to hear your opinion about that. I'll implement analogous to [#112](https://github.com/imaNNeo/fl_chart/pull/112) and add tests :)

imaNNeo:
I appreciate it!
Thanks in advance.

jay-k98:
Hi @imaNNeo i was on vacation but now I'm back and implemented the stacked logic for PieChart. I also updated tests and added a new sample for the example app. Looking forward to your feedback :)

Bazni:
Hey 👋,
I think I commented at the wrong place the first time, sorry about that.

Thx for the work. I'm looking to do something like this:
image


Do you think I will be able to do it with your StackedPieChart?

imaNNeo:
> Hi @imaNNeo i was on vacation but now I'm back and implemented the stacked logic for PieChart. I also updated tests and added a new sample for the example app. Looking forward to your feedback :)

Can you please update your branch with the latest changes? Then I can check it again.
(You don't need to update the CHANGELOG.md)

jay-k98:
> Can you please update your branch with the latest changes? Then I can check it again. (You don't need to update the CHANGELOG.md)

Hey, I updated the branch and removed CHANGELOG.md changes.

jay-k98:
> Hey 👋, I think I commented at the wrong place the first time, sorry about that.
> 
> Thx for the work. I'm looking to do something like this: image
> 
> Do you think I will be able to do it with your StackedPieChart?

Yes, the green and red part will definetly be possible. You can checkout the branch of this PR and run the example. Pie Chart Sample 4 is pretty close to the screenshot you provided.

About the grid you might need to get creative.. :D But I guess you could put two PieCharts in a Stack and use the "lower" PieChart to create the grid with section and segment outlines as well as labels. Hope this helps :)

imaNNeo:
Can you please update your branch?

jay-k98:
@imaNNeo Thank you for responding. I updated the branch and applied the suggested changes of the copilot code-review.

imaNNeo:
Thanks for the fixes,
I had a closer look now. And there's something that doesn't feel right to me.

We have the `PieChartSectionData.radius` and also segments (where each segment has `PieChartStackSegmentData.radius`

Now they're adding to determine the final radius of the pie-section shape.
But what I was assuming is like:
`PieChartSectionData.radius` -> determines the total radius of the section
Then we can have a list of:
`PieChartStackSegmentData` -> that each one determines the from/to radius of the stacked segment.

Just like how stacking works in BarChart (Look at BarChartSample4 for example)
image
In BarChart, we define the main bar's `toY`, then each stack item can have a `fromY` and `toY`. We render the main bar behind, then we render the stacked parts on top of that based on the `fromY` and `toY`

I would like to have a similar functionality here, so in that way, we can have `PieChartSectionData.radius` which determines the total radius of the section, and we render it behind, then we will have a list of segments that each one has a `fromRadius` and `toRadius`, then we render them on top of the base section.
(Keep in mind to clamp the numbers based on the valid radius that we can have (0 to PieChartSectionData.radius)

So basically, `segments` is a list that we render on top of the main section!

This way:
1. We have the consistent logic for our stacked items in different charts
2. We have more customizability (for example user can add a stack only in the middle of the main section

Sorry for giving this feedback so late, but this is something I realized now!
Please let me know your thoughts on it.

jay-k98:
@imaNNeo I get your point and I like the idea of keeping it consistent. Also as you say it would lead to a more flexible solution, which I also appreciate. I will start refactoring and hit you up when I'm done.

jay-k98:
@imaNNeo I updated the branch and commited the refactored segment logic. I really like it because it is much cleaner and easier to understand. Thanks a lot for your feedback! Segment spacing can be solved by adding a segment. I provided that trick in the samples.

Let me know if this is what you had in mind :)

imaNNeo:
Nice! Thanks for the changes. I also like the new structure. Just some minor comments to solve before merging it:

lib/src/chart/pie_chart/pie_chart_painter.dart
```diff
@@ -175,12 +154,134 @@ class PieChartPainter extends BaseChartPainter<PieChartData> {
...
+    CanvasWrapper canvasWrapper,
+    PieChartSectionData section,
+    double sweepAngle,
+    double startRadius,
+    double startAngle,
+    Offset center,
+  ) {
+    if (section.radius <= 0) return;
+
+    final mainPath = generateSegmentPath(
```
imaNNeo: As I understood, we already have this path created at line 148 (called `mainPath`) in this file (the caller of this function).
So I think it's better to pass the already created path into this function and reuse it. Instead of re-creating the same path again. Please correct me if I'm wrong.

imaNNeo (changes requested):

imaNNeo:
And please:
* Add some unit-tests for the added classes and the functionality
* Add the new property/feature in our documentation in pie_chart.md

After that, we're ready to go!

jay-k98:
@imaNNeo Thanks for your feedback. I implemented reusing the path and added some tests for better test coverage as well as documentation. Ready for your approval :)

imaNNeo (approved):

imaNNeo:
Perfect! 
Thanks for your contribution!

PR #2007, opened 2025-09-25
Feat/radar max value config

artsooter:
hello ，This PR is aims to add this feature in https://github.com/imaNNeo/fl_chart/issues/2002

Problem background
---
We encountered two problems in the issue.

* How to set the maximum value of the Radar Chart to a specified value.
* How to display the maximum tick value on the chart.

Solution
---
This PR introduces two new parameters to RadarChartData:

1. maxValue Parameter

  `RadarChartData(
    dataSets: myDataSets,
    maxValue: 100.0,  // Set custom maximum value
  )`
  - Type: double? (optional)
  - Default: null (uses data maximum)
  - Purpose: Allows setting a custom maximum value instead of calculating from data

2. showMaxTick Parameter

 ` RadarChartData(
    dataSets: myDataSets,
    showMaxTick: true,  // Show maximum tick value
  )`
  - Type: bool
  - Default: false (preserves existing behavior)
  - Purpose: Controls whether the maximum tick is displayed on the chart

Result
---
image

### DIFF
--- a/lib/src/chart/radar_chart/radar_chart_data.dart
+++ b/lib/src/chart/radar_chart/radar_chart_data.dart
@@ -76,6 +76,7 @@ class RadarChartData extends BaseChartData with EquatableMixin {
     RadarTouchData? radarTouchData,
     this.isMinValueAtCenter = false,
     this.maxValue,
+    this.showMaxTick = false,
     super.borderData,
   })  : assert(dataSets != null && dataSets.hasEqualDataEntriesLength),
         assert(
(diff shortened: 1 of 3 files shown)


imaNNeo:
Hi, thanks for your great contribution!
I have some points to mention.

1. I think instead of having a `showMaxTick`, we can have a `checkToShowTick` resolver that the user can decide to show which ticks (it can be like a callback, like `checkToShowDot` in LineChartData). This way, we can hide the max entry by default (to be backward compatible), also it adds a lot of customizability for other use cases. 
2. I like the change for the `maxValue` property, thanks a lot. I was thinking to have something for `minValue` as well (but not for now, let's keep this PR as simple as possible). But if you found time in the future, I would appreciate if you add the `minValue` as well to keep the consistency.
3. Please add the newly added property in the `radar_chart.md` file.

Then we're good to go!

artshooter:
Hello , I sincerely apologize for the inconvenience, but the author account of this PR has been retired and can no longer be used.

Therefore, I have submitted a new PR using my new account, which implements the features based on the feedback. https://github.com/imaNNeo/fl_chart/pull/2025

Thank you.

Issue #2002, opened 2025-09-22
How to set the maximum value for Radar Chart in FL Chart


HaroldVincent1408:
I’m using FL Chart version 1.1.0, and I’m currently working with a Radar Chart. However, the ticks on my chart are not displaying in a nice way. I’ve checked issues #995 and #1091 , but I’m still not sure if the feature to set the maxValue for Radar Chart has been implemented yet, and how to properly use it.
Here's my chart, I would like the ticks to be displayed as 25, 50, 75, and the largest value as 100.
Any help would be greatly appreciated!
<img width="268" height="278" alt="Image" src="https://github.com/user-attachments/assets/63ff2841-8d13-422c-83dd-3284eb366250" />

artsooter:
Hello, I think your question can be broken down into two parts:

1. How to set the maximum value of the Radar Chart to 100.
2. How to display the maximum tick value on the chart.

--- 

**For Question 1:**

The current logic in the code determines the maximum value of the Radar Chart based on the largest data point value, and it does not support manually setting a maximum value. 

<img  width="402" alt="Image" src="https://github.com/user-attachments/assets/c1d3d950-d6fc-4d4a-bdbe-a6f396cc37b9" />

To achieve a maximum value of 100, maybe you could try adding a transparent `RadarDataSet` with a data point value of 100 to wrap the entire chart, effectively setting the chart's maximum value to 100.

just like this

<img width="300"  alt="Image" src="https://github.com/user-attachments/assets/920ca043-2b67-4cc7-8376-c8bb20149bee" />

<img width="406"   alt="Image" src="https://github.com/user-attachments/assets/f4b5b0f6-5f81-4658-89fc-a3f94ee9b604" />


--- 

**For Question 2:**
The current code logic displays `tickCount - 1` tick marks on the chart. Unfortunately, I haven’t found a straightforward way to achieve the desired effect of displaying specific tick values (e.g., 25, 50, 75, 100) without modifying the source code.

<img width="400" alt="Image" src="https://github.com/user-attachments/assets/0de3642c-68a8-47ff-bb89-f3f863beeabd" />

HaroldVincent1408:
Thanks again for your support. You really pointed me in the right direction. Have a great day sir

PR #2025, opened 2025-10-27
feat: add maxValue and checkToShowTick in radarChart

artshooter:
## Summary

  This PR implements the customization features for RadarChart as discussed in #2002.

  **Note**: This is a follow-up to PR #2007. The previous PR was submitted from the GitHub account `artsooter`, which is no longer available. 
  I've created this new PR with a fresh account to implement the feedback.

  ## Changes

  1. **maxValue parameter**: Allows setting a custom maximum value for the RadarChart instead of auto-calculating from data.

  2. **checkToShowTick callback**: Implemented as suggested in the review. This function determines which ticks should be displayed by receiving the tick
  index,  and list of all ticks. Returns a boolean to control visibility. Default behavior shows all ticks except the last one.

  3. **Documentation**: Added descriptions for both parameters to `radar_chart.md`.

  ## Documentation

  |PropName|Description|default value|
  |:-------|:----------|:------------|
  |maxValue|Defines a custom maximum value for the [RadarChart]. If provided, this value will be used instead of the automatically calculated maximum from the data.|null|
  |checkToShowTick|Determines which ticks should be displayed on the [RadarChart]. By default, shows all ticks except the last one.|showTickWithoutLast|

  ---

### DIFF
--- a/lib/src/chart/radar_chart/radar_chart_data.dart
+++ b/lib/src/chart/radar_chart/radar_chart_data.dart
@@ -12,10 +12,19 @@ typedef GetTitleByIndexFunction = RadarChartTitle Function(
   double angle,
 );
 
+typedef CheckToShowTick = bool Function(
+  int index,
+  List<double> ticks,
+);
+
 enum RadarShape {
   circle,
   polygon,
 }
+
+/// Shows all ticks except the last one.
+bool showTickWithoutLast(int index, List<double> ticks) =>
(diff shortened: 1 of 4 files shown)


XOYZ69:
Is there any update if this gets implemented / merged in the near future? Would definitely make my life a bit simpler if this PR gets approved and added to the next release.

Issue #2027, opened 2025-11-01
Feature Request - Draggable Points


andy-edvalson:
So far super impressed with this library!

## Problem
FL Chart currently doesn't support dragging data points, which limits its use 
for interactive data editing applications (scheduling, profile editing, etc.)

## Proposed Solution
Add optional drag-and-drop functionality to LineChart that allows users to:
- Select data points by tapping
- Drag selected points to new positions
- Receive callbacks with updated point positions
- [Optional] Validate moves via callbacks (e.g., prevent out-of-bounds updates)

## Use Case
I've implemented this in a pretty proprietary fashion, including gesture detection, throttling, nearest point detection, and final position capture for a production app and might be able to contribute it back. 

Would it be helpful if I attempt to adapt it to a PR?

PR #2028, opened 2025-11-01
[#2027] Add DraggableLineChart widget for drag-to-edit functionality

andy-edvalson:
Adds a wrapper class to LineChart which allows dragging of points.

Does not support existing in scrollable contexts, and does not play super nice with the tooltip support, so maybe someone cleverer than I could implement this a little bit closer to LineChart without requiring the wrapper class.

This approach worked great for a timing scheduler for us, so I wanted to offer this up in case it's helpful.


andy-edvalson:
@imaNNeo Would you like to look over this, If not I can just use my own fork

imaNNeo:
Feel free to use your fork now. 
I'll review it when I have some time. 
Thanks!

PR #2032, opened 2025-11-13
feat: candlestick chart support CandleMarkerPainter

artshooter:
Hi, I've made improvements and refinements based on the feedback from the previous PR.

---

  Feature Summary

  This PR adds a candle marker system to CandlestickChart for marking trading signals on candlestick charts.

---

  Key Features:

  - CandleMarkerPainter - Abstract base class for custom candle marker implementations
  - BuyAndSellSignalPainter - Built-in implementation with flexible condition-based signal detection,
  customizable bubble styles and labels to easily extend various trading signal scenarios

### DIFF
--- a/lib/src/chart/candlestick_chart/candlestick_chart_data.dart
+++ b/lib/src/chart/candlestick_chart/candlestick_chart_data.dart
@@ -50,6 +50,7 @@ class CandlestickChartData extends AxisChartData with EquatableMixin {
     super.backgroundColor,
     super.rotationQuarterTurns,
     this.touchedPointIndicator,
+    this.candleMarkerPainter,
   })  : candlestickSpots = candlestickSpots ?? const [],
         candlestickPainter = candlestickPainter ?? DefaultCandlestickPainter(),
         candlestickTouchData = candlestickTouchData ?? CandlestickTouchData(),
@@ -108,6 +109,13 @@ class CandlestickChartData extends AxisChartData with EquatableMixin {
   /// if you want to have customized [touchedPointIndicator]
   final AxisSpotIndicator? touchedPointIndicator;
 
+  /// The painter used to draw markers on candlesticks.
+  ///
+  /// For example, buy/sell signal markers, important point markers, etc.
+  /// If null, no markers will be drawn.
+  /// You can use the built-in [BuySellSignalMarkerPainter] or implement your own.
+  final CandleMarkerPainter? candleMarkerPainter;
+
   /// Lerps a [CandlestickChartData] based on [t] value, check [Tween.lerp].
   @override
   CandlestickChartData lerp(BaseChartData a, BaseChartData b, double t) {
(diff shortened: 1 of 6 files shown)


SAGARSURI:
Can I draw a line to show trade indicators? Something like this:

artshooter:
@SAGARSURI 
Unfortunately, this PR implements buy/sell signal markers on individual candles, not horizontal lines across the chart. The screenshot shows an order line feature that displays price lines, position size, and P&L information. These are fundamentally different features, which are not currently supported.

lib/src/chart/candlestick_chart/candlestick_chart_data.dart
```diff
@@ -997,3 +1009,256 @@ class CandlestickChartDataTween extends Tween<CandlestickChartData> {
...
+abstract class CandleMarkerPainter with EquatableMixin {
+  const CandleMarkerPainter();
+
+  /// Determines whether this marker should be shown on the given candle spot.
+  ///
+  /// [spot] contains the candlestick data (open, high, low, close, x).
+  /// [spotIndex] is the index of the spot in the candlestickSpots list.
+  ///
+  /// Return true if you want to draw this marker on this spot, false otherwise.
+  bool shouldShow(CandlestickSpot spot, int spotIndex);
```
imaNNeo, line 1025: Do we actually need it? I think we can have an empty paint() implementation to not show anything.
This way simplifies the implementation

lib/src/chart/candlestick_chart/candlestick_chart_data.dart
```diff
@@ -997,3 +1009,256 @@ class CandlestickChartDataTween extends Tween<CandlestickChartData> {
...
+    CandleMarkerPainter b,
+    double t,
+  );
+
+  @override
+  List<Object?> get props => [];
+}
+
+/// Signal type enum for buy/sell markers.
+enum BuyAndSellSignalType {
```
imaNNeo, line 1056: In my opinion, we should prevent defining domain-specific models inside the fl_chart. And it's better to keep things generic.

So I suggest removing it (and adding it inside the candlestick_chart_sample1.dart if you want), and here we can rename the BuyAndSellSignalPainter() to CandleBalloonTextPainter(). This way, it feels like a generic implementation of balloon text, and you can use it to show buy/sell in your implementation. 

And the logic of CandleBalloonTextPainter() will be way simpler, and the complexity goes into the candlestick_chart_sample1.dart where you want to show sell/buy points.

lib/src/chart/candlestick_chart/candlestick_chart_data.dart
```diff
@@ -108,6 +109,13 @@ class CandlestickChartData extends AxisChartData with EquatableMixin {
   /// if you want to have customized [touchedPointIndicator]
   final AxisSpotIndicator? touchedPointIndicator;
 
+  /// The painter used to draw markers on candlesticks.
+  ///
+  /// For example, buy/sell signal markers, important point markers, etc.
+  /// If null, no markers will be drawn.
+  /// You can use the built-in [BuySellSignalMarkerPainter] or implement your own.
+  final CandleMarkerPainter? candleMarkerPainter;
```
imaNNeo, line 117: What do you think if we allow users to have a list of CandleMarkerPainter? 
This way, we can let the user compose different marker painters to render on top of each other.

imaNNeo (changes requested):

PR #2024, opened 2025-10-24
feat: pie chart rounded corners

Vizten18:
fixes: https://github.com/imaNNeo/fl_chart/issues/1175

## Summary
Add the ability to apply rounded corners to individual pie chart sections using a simple `cornerRadius` parameter.

## Motivation
Currently, fl_chart pie charts only support sharp corners for sections. Many modern UI designs require rounded corners for a softer, more polished appearance. This feature would allow developers to create more visually appealing pie charts that match contemporary design trends.

## Proposed Solution
Add a `cornerRadius` property to `PieChartSectionData` that allows developers to specify the radius for rounding the corners of individual pie chart sections.

## Expected API
```dart
PieChartSectionData(
  value: 25,
  color: Colors.blue,
  cornerRadius: 8.0, // New parameter
  // ... other existing parameters
)

##  Visual Improvements
- Smoother, more modern appearance
- Consistent with Material Design 3 principles
- Eliminates sharp edges for better visual hierarchy

##  Documentation
- Comprehensive method documentation with behavior notes
- Inline comments explaining clamping logic and geometric calculations
- Platform compatibility notes (web-html renderer caveats)

PR #2040, opened 2025-12-14
Pie chart section border radius support

Mellow-Developer:
This commit adds the support for border radius on pie chart sections.
There is a limitation for now that border radius must be only in a range of 0-8px. otherwise it throws exception.


imaNNeo:
I don't think throwing an exception is a good idea; it's better to handle it gracefully (maybe by clamping it)
But, there's another open PR for this topic here:
#2024

So let's go ahead with that one. I'll mention your PR there, so we can get idea from.
Thanks for your contribution!

PR #2049, opened 2026-01-25
feat: Add FlDotImagePainter for custom image dot markers

### LINKED ISSUE #2050: [Feature Request] Add FlDotImagePainter to support custom image dot markers
**Is your feature request related to a problem? Please describe.**

Currently, fl_chart only supports basic geometric shapes (circle, square, cross) for dot markers on line charts through `FlDotCirclePainter`, `FlDotSquarePainter`, and `FlDotCrossPainter`. 

However, many use cases require custom visual markers such as:
- User profile pictures on activity charts
- Product icons on sales trend charts  
- Custom branded icons for different data series
- Weather icons on temperature charts

There's no built-in way to use custom images as dot markers, forcing developers to either:
1. Use the limited geometric shapes
2. Implement complex custom painters from scratch
3. Use workarounds that don't integrate well with the existing dot system

I propose adding a `FlDotImagePainter` class that extends `FlDotPainter` to support rendering custom images as dot markers.

**Proposed API:**
```dart
// Load image asynchronously
final image = await FlDotImagePainter.loadImageFromAsset('assets/icon.png');
final painter = FlDotImagePainter(image: image, size: 20.0);

// Use in chart
LineChart(
  LineChartData(
    lineBarsData: [
      LineChartBarData(
        dotData: FlDotData(
          show: true,
          getDotPainter: (spot, percent, barData, index) => painter,
        ),
      ),
    ],
  ),
)

### PR DESCRIPTION
**Pull request by InsoooooooooJANG**
# Description
Adds `FlDotImagePainter`, an implementation of `FlDotPainter` that draws a `dart:ui` `Image` as the dot marker, instead of the built-in circle, square or cross shapes.

The image is centered on the spot and scaled to fit inside a `size` x `size` square, keeping its aspect ratio. Since `FlDotPainter.draw()` is synchronous, the image has to be decoded before the painter is created — loading is left to the user (asset, network, generated), and `LineChartSample14` shows one way to do it.

```dart
final data = await rootBundle.load('assets/dot.png');
final buffer = await ui.ImmutableBuffer.fromUint8List(data.buffer.asUint8List());
final codec = await PaintingBinding.instance.instantiateImageCodecWithSize(buffer);
final frame = await codec.getNextFrame();

final painter = FlDotImagePainter(image: frame.image, size: 20);

LineChartBarData(
  spots: spots,
  dotData: FlDotData(
    getDotPainter: (spot, percent, barData, index) => painter,
  ),
)
```

image

## Checklist

- [x] I have updated/added relevant documentation and added dartdoc comments with `///`.
- [x] I have updated/added relevant examples in `example`.

## Breaking Change?

## Related Issues

Closes #2050


lib/src/chart/base/axis_chart/axis_chart_data.dart
```diff
@@ -1466,6 +1467,73 @@ abstract class FlDotPainter with EquatableMixin {
...
+  final double size;
+
+  /// Loads an image from asset path and returns a [Image] object.
+  ///
+  /// Example:
+  /// ```dart
+  /// final image = await FlDotImagePainter.loadImageFromAsset('assets/dot.png');
+  /// final painter = FlDotImagePainter(image: image, size: 20.0);
+  /// ```
+  static Future<Image> loadImageFromAsset(String assetPath) async {
```
imaNNeo, line 1495: loadImageFromAsset doesn't belong in the painter.
A static method on a painter that does rootBundle.load() and instantiateImageCodec() mixes asset loading with rendering. This should be the user's responsibility.
They load the image however they want (asset, network, generated) and pass the Image object in. The static helper could live in an example or utility, but not in the library's core data class.

imaNNeo: btw, it seems instantiateImageCodec() is deprecated in Flutter 3.x.
I get this warning as well:
Warning: In index.html:92: "FlutterLoader.loadEntrypoint" is deprecated. Use "FlutterLoader.load" instead. See https://docs.flutter.dev/platform-integration/web/initialization for more details.

So please make sure you're using the new approach to load the image

lib/src/chart/base/axis_chart/axis_chart_data.dart
```diff
@@ -1466,6 +1467,73 @@ abstract class FlDotPainter with EquatableMixin {
...
+    final frame = await codec.getNextFrame();
+    return frame.image;
+  }
+
+  @override
+  void draw(Canvas canvas, FlSpot spot, Offset offsetInCanvas) {
+    // Center the image at the offset
+    final drawOffset = offsetInCanvas - Offset(size / 2, size / 2);
+    final rect = Rect.fromLTWH(drawOffset.dx, drawOffset.dy, size, size);
+    paintImage(
```
imaNNeo, line 1507: Please use canvasWrapper.drawImage() instead of paintImage. This way, we keep the consistency and testability. Just like here:

https://github.com/imaNNeo/fl_chart/blob/c83aa9049d26b1940e9f4d7ddd1b578b2711a94f/lib/src/chart/base/axis_chart/axis_chart_painter.dart#L296-L305

example/lib/presentation/samples/line/line_chart_sample14.dart
```diff
@@ -0,0 +1,122 @@
...
+      child: Padding(
+        padding: const EdgeInsets.all(16),
+        child: LineChart(
+          LineChartData(
+            // Define the chart boundaries
+            minX: 0,
+            maxX: 6,
+            minY: 0,
+            maxY: 6,
+            // Configure axis titles
```
imaNNeo, line 60: I think we don't need these kinds of comments. Please remove the unnecessary comments.

imaNNeo (changes requested):

imaNNeo:
You also need to write some unit-tests in the end.

lib/src/chart/base/axis_chart/axis_chart_data.dart
InsoooooooooJANG: Done — `loadImageFromAsset()` is gone from the library. `FlDotImagePainter` now only accepts an already-decoded `Image`, and `axis_chart_data.dart` no longer imports `flutter/services.dart` (it was only there for `rootBundle`). Loading moved into `LineChartSample14`.

On the deprecation: I double-checked and `ui.instantiateImageCodec` isn't actually deprecated — there's no `@Deprecated` annotation on it in either Flutter 3.38 or 3.44. The `FlutterLoader.loadEntrypoint` warning you saw is a separate, pre-existing thing coming from `example/web/index.html:92`; it shows up on `main` too, unrelated to this PR. I left it alone since it's out of scope here — happy to send a small separate PR for it if you want.

That said, the sample now sidesteps the question entirely. It uses:

```dart
final data = await rootBundle.load('assets/icons/image_annotation.png');
final buffer = await ui.ImmutableBuffer.fromUint8List(data.buffer.asUint8List());
final codec = await PaintingBinding.instance.instantiateImageCodecWithSize(buffer);
final frame = await codec.getNextFrame();
```

`instantiateImageCodecWithSize` is the documented replacement for `instantiateImageCodecFromBuffer` (that one *is* deprecated), and it disposes the buffer for you.

One thing worth knowing if you ever reach for it: I first wrote this with `ui.ImmutableBuffer.fromAsset()`, which is shorter, but it throws `UnsupportedError` on web — the sample just spun forever on the web demo. Going through `rootBundle` + `fromUint8List` works everywhere.

example/lib/presentation/samples/line/line_chart_sample14.dart
InsoooooooooJANG: Removed. The sample now reads like the other line samples — no class doc header, no comments labelling parameters. The only one left is the line explaining *why* the image has to be decoded up front, which didn't seem obvious from the code alone.

While I was in there I also aligned it with the surrounding samples: it uses `AppColors` instead of raw `Colors`, guards `setState` with `if (!mounted)` after the `await`, and keeps the layout stable while loading (like `LineChartSample13`) instead of returning a bare spinner that made the card resize once the image landed.

lib/src/chart/base/axis_chart/axis_chart_data.dart
InsoooooooooJANG: `paintImage` is gone. I couldn't route it through `CanvasWrapper` though, and I'd rather check the direction with you than guess:

- `FlDotPainter.draw` only receives a raw `Canvas` — `CanvasWrapper.drawDot` unwraps it at [`canvas_wrapper.dart:123`](https://github.com/imaNNeo/fl_chart/blob/main/lib/src/utils/canvas_wrapper.dart#L122-L124), and that's the only call site.
- `CanvasWrapper` isn't exported from `lib/fl_chart.dart`, so changing `draw` to take one would make `FlDotPainter` impossible for users to implement without importing `package:fl_chart/src/...`. That's a public breaking change.
- `Canvas.drawImage` (the one in the code you linked) has no scaling — it draws at native size. Using it directly would silently ignore `FlDotImagePainter.size`.
- The raw-`Canvas` shape also seems to be the convention for user-extensible painters: `GaugeTickPainter`, `GaugeMarkerPainter` and `GaugePointerPainter` all take a `Canvas` and call `canvas.drawLine`/`save`/`translate` directly, as do the three built-in dot painters and `FlSimpleErrorPainter`.

So I dropped `paintImage` for `canvas.drawImageRect` with `applyBoxFit(BoxFit.contain, ...)`, which keeps both `size` and the aspect ratio working. For the testability half of your point: `draw` is now unit-tested against `MockCanvas` and asserts the exact source/destination rects, the same way `FlSimpleErrorPainter.draw` is tested in `axis_chart_data_test.dart`.

If you'd rather have `CanvasWrapper` in the painter interfaces, I'm glad to do it — I just think it deserves its own breaking PR that exports `canvas_wrapper.dart` and converts all five painter interfaces at once, rather than changing only this one. Let me know which you prefer.

InsoooooooooJANG:
Thanks for the review, and sorry for the long silence on this one. All four points are addressed and `main` is merged in, so the conflict is gone.

**Unit tests** — added a `FlDotImagePainter` group to `test/chart/base/axis_chart/axis_chart_data_test.dart` covering equality, `getSize`, `mainColor`, same-type and cross-type `lerp`, the inherited `hitTest`, and `draw` (asserting the exact src/dst rects that come out of `applyBoxFit`, plus that a square image fills the whole dot). `MockCanvas` was already generated for that file, so no `make codeGen` needed.

**One change beyond the review**: `mainColor` was returning `Colors.transparent`, which would render scatter-chart tooltip text invisible (it feeds the text color at `scatter_chart_data.dart:577`). It's now a constructor field defaulting to `Colors.green`, matching `FlDotCirclePainter.color`.

I also dropped `example/assets/icons/ic_meal.png` — the sample loads the existing `image_annotation.png`, so that file was never referenced.

Verified with `make runTests` (673 passing) and `make checkFormat`, and I ran the example on web to confirm the sample actually renders.

**Heads-up on CI**: `make analyze` is currently failing on `main` itself — the run for f3a5758 is red with 167 pre-existing `EquatableMixin is deprecated` infos coming from a newer `equatable` resolving on current stable. So the red check here isn't from these changes. Want me to send a separate PR that switches those over to `Equatable`?

PR #2066, opened 2026-03-13
fix!: Remove non-functional borderData from PieChartData

### LINKED ISSUE #1182: PieChartData->borderData not working
PieChartData->borderData not working

My Code
```
PieChartData(
	pieTouchData: PieTouchData(
	  // Touch Data Code
	),
	borderData: FlBorderData(show:true),
	sections: List.generate(5,(i){
	  // Section Code
	})
),
```
Result:
Screenshot_1

Flutter - 3.3.6
FlChart - 0.55.1

mahmoodhamdi:
## Problem

`PieChartData` inherits `borderData` from `BaseChartData`, but `PieChartPainter` never uses it to draw any borders. When users set `borderData: FlBorderData(show: true)` on a `PieChartData`, nothing happens — the parameter is silently ignored.

This was reported in #1182 and confirmed by the maintainer.

## Solution

Removed the `borderData` parameter from `PieChartData`:
- Removed from constructor, `copyWith`, `lerp`, and `props`
- `PieChartData` now always passes `FlBorderData(show: false)` to `BaseChartData`
- Updated documentation, example app, and tests

Users should use `PieChartSectionData.borderSide` for section borders.

## Breaking Change

Code that passes `borderData` to `PieChartData(...)` will get a compile error. The fix is to simply remove the `borderData` parameter — it was never functional.

**Migration:**
```dart
// Before
PieChartData(
  borderData: FlBorderData(show: false), // remove this
  sections: [...],
);

// After
PieChartData(
  sections: [...],
);
```

## Testing

- All 573 existing tests pass
- `flutter analyze` — zero warnings
- `make checkstyle` — passes
- Removed tests that validated `borderData` equality on `PieChartData`

## Checklist
- [x] `flutter test` passes (573 tests)
- [x] `dart format .` applied
- [x] CHANGELOG.md updated

Closes #1182


imaNNeo:
Comment from [here](https://github.com/imaNNeo/fl_chart/issues/1182#issuecomment-4223463892):

> Removing it will cause a breaking change for users who are using this property (even though it doesn't do anything).
> 
> On the other hand, I think there might be a use case where users want to have a rectangular border around the chart, and they can use this property. This way, we keep it consistent with the other chart types (as we have it intentionally in our BaseChartData).
> 
> I'm sorry, but I was wrong in my [previous comment](https://github.com/imaNNeo/fl_chart/issues/1182#issuecomment-1383156358). Now I'm seeing it from a different perspective. Let's keep it and make it functional.

PR #2067, opened 2026-03-13
fix: Correct touched section index in PieChart with zero-value sections

### LINKED ISSUE #1632: PieChart: touched section index is correct if some values are zero
PieChart: touched section index is incorrect if some values are zero. For example,

`    values: 0, 0, 100, 200, 300.`

when touching the 3rd value 100, touched index is 0 that should be 2.

mahmoodhamdi:
## Problem

When a `PieChart` has sections with zero values (e.g., `[0, 0, 100, 200, 300]`), touching any visible section returns the wrong `touchedSectionIndex`. For example, touching the section with value `100` (index 2) incorrectly returns index `0`.

## Root Cause

The `drawSections` method skips zero-value sections with `continue`, so they aren't drawn. However, `handleTouch` does not skip them — it generates a section path with a zero sweep angle, which can incorrectly match a touch point before the actual visible section is checked.

## Solution

Added the same `if (section.value == 0) continue;` check to the `handleTouch` loop, matching the behavior of `drawSections`.

## Testing

Added a unit test with sections `[0, 0, 100, 200, 300]` that verifies:
- Touching section with value 100 returns index 2
- Touching section with value 200 returns index 3
- Touching section with value 300 returns index 4
- Touching center returns -1

## Checklist
- [x] `flutter test` passes
- [x] `dart format .` applied
- [x] CHANGELOG.md updated

Closes #1632


CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 ## newVersion
+* **BUGFIX** (by @mahmoodhamdi) Fix incorrect touched section index in PieChart when some sections have zero values, #1632
```
imaNNeo, line 2: You don't need to add the changelog line in this file anymore.
We'll generate it based on the commit history from now on.

imaNNeo (changes requested):

imaNNeo:
Good catch!
It's ready to merge. 
Please remove your changelog in CHANGELOG.md, and update your branch. 
Then we're good to go!

PR #2068, opened 2026-03-13
fix: Clip only specified sides when using FlClipData

### LINKED ISSUE #1262: Whenever any clip side is true, all sides are clipped
Hi! First of all, thanks for this great library. The amount of options to customize are incredible. 

**Describe the bug**
It looks like clipping is applied to all sides of a line graph even when it is enabled on just one side.

**To Reproduce**
- Open the example app (Line Chart Sample 10)
- Set all parameters of `FlClipData` to `false`
  - Note that there is *no* clipping on the right side of the graph
- Set any parameter (except `right`) of `FlClipData` to `true`
  - Note that there *is* clipping on the right side of the graph

**Screenshots**
image

image

**Versions**
 - Flutter 3.7.3
 - FlChart 0.60.0

mahmoodhamdi:
## Problem

When any side of `FlClipData` is enabled (e.g., `FlClipData(top: true, bottom: false, left: false, right: false)`), all sides get clipped — not just the specified one. Content that extends beyond the view on non-clipped sides is incorrectly hidden.

## Root Cause

Both `saveLayer` and `clipRect` used the exact view bounds as their rect boundaries. Non-clipped sides defaulted to the view edges (e.g., `right = viewSize.width`), which prevented content from overflowing even when clipping was disabled on that side.

## Solution

For non-clipped sides, the clip rect boundaries now extend well beyond the view (using the sum of width and height as a generous overflow buffer). This allows content to overflow freely on non-clipped sides while still respecting clipping on the specified sides.

The same fix was applied to all three affected painters:
- `LineChartPainter.clipToBorder()`
- `ScatterChartPainter.drawSpots()`
- `CandlestickChartPainter.drawCandlesticks()`

## Testing

- Updated existing `clipToBorder` tests to verify that non-clipped sides use extended boundaries
- All 574 existing tests pass
- `make sure` passes (tests + checkstyle)

## Checklist
- [x] `flutter test` passes
- [x] `dart format .` applied
- [x] CHANGELOG.md updated

Closes #1262


mahmoodhamdi:
Closing — apologies for the noise.

Issue #517, opened 2020-12-16
Outline the graph line


subzero911:
Please add the `outline` parameter for LineChartBarData.
It should have `strokeWidth` and `strokeColor` properties.

So I want be able to draw such a line
![image](https://user-images.githubusercontent.com/12999702/102374607-96c35d80-3fd2-11eb-8c9a-ae45f976effc.png)
(Sorry for my exceptional drawing skills)
And maybe also separate `outlineAbove: true, outlineBelow: true`, and `capEnds: false`

PR #2076, opened 2026-03-22
feat: Add stroke support for LineChartBarData

patilrahu:
## Description

This PR adds optional stroke
Simulator Screenshot - iPhone 16e - 2026-03-22 at 15 22 34
 support for LineChartBarData.

### Changes
- Added `strokeColor` and `strokeWidth` properties
- Implemented border rendering behind the main line in the painter
- Maintains backward compatibility (default = no border)
- Updated example to demonstrate usage

This feature improves line visibility, especially when multiple lines overlap or when used on complex backgrounds.

Closes Outline the graph line #517

### DIFF
--- a/lib/src/chart/line_chart/line_chart_data.dart
+++ b/lib/src/chart/line_chart/line_chart_data.dart
@@ -251,6 +251,8 @@ class LineChartBarData with EquatableMixin {
     this.isStrokeJoinRound = false,
     BarAreaData? belowBarData,
     BarAreaData? aboveBarData,
+    this.strokeColor,
+    this.strokeWidth = 0,
     this.dotData = const FlDotData(),
     this.errorIndicatorData =
         const FlErrorIndicatorData<LineChartSpotErrorRangeCallbackInput>(),
@@ -346,6 +348,14 @@ class LineChartBarData with EquatableMixin {
   /// otherwise it draws line with hard edges.
   final bool isCurved;
 
+  /// Optional border color for the line.
+  /// Useful for improving visibility when lines overlap.
+  final Color? strokeColor;
+
+  /// Width of the border around the line.
+  /// Default is 0 (no border).
+  final double strokeWidth;
+
   /// If [isCurved] is true, it determines smoothness of the curved edges.
   final double curveSmoothness;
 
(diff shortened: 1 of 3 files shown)


lib/src/chart/line_chart/line_chart_data.dart
```diff
@@ -346,6 +348,14 @@ class LineChartBarData with EquatableMixin {
   /// otherwise it draws line with hard edges.
   final bool isCurved;
 
+  /// Optional border color for the line.
+  /// Useful for improving visibility when lines overlap.
+  final Color? strokeColor;
```
imaNNeo, line 353: Can we use the [BorderSide](https://api.flutter.dev/flutter/painting/BorderSide-class.html) class instead of defining strokeColor and strokeWidth here?
Because all other chart types in fl_chart use BorderSide for borders (e.g., BarChartRodData.borderSide, PieChartSectionData.borderSide). 

Let's keep the API consistent:
```dart
final BorderSide borderSide; // default: BorderSide.none
```

lib/src/chart/line_chart/line_chart_painter.dart
```diff
@@ -281,6 +281,18 @@ class LineChartPainter extends AxisChartPainter<LineChartData> {
...
+        final borderPaint = Paint()
+          ..color = barData.strokeColor!
+          ..strokeWidth = barData.barWidth + (barData.strokeWidth * 2)
+          ..style = PaintingStyle.stroke
+          ..strokeCap =
+              barData.isStrokeCapRound ? StrokeCap.round : StrokeCap.butt
+          ..strokeJoin =
+              barData.isStrokeJoinRound ? StrokeJoin.round : StrokeJoin.miter;
+
+        canvasWrapper.canvas.drawPath(barPath, borderPaint);
```
imaNNeo, line 294: This line bypasses the canvas wrapper. All drawing must go through canvasWrapper.drawPath(...)

So please use `.drawPath()` directly on the `canvasWrapper`

lib/src/chart/line_chart/line_chart_painter.dart
```diff
@@ -281,6 +281,18 @@ class LineChartPainter extends AxisChartPainter<LineChartData> {
         barData,
       );
       drawBarShadow(canvasWrapper, barPath, barData);
+      if (barData.strokeColor != null && barData.strokeWidth > 0) {
+        final borderPaint = Paint()
```
imaNNeo, line 285: There's a performance issue with creating the Paint() object in every frame.
Please create the Paint instance once (in the constructor, just like other Paint() objects) and reuse it here.

imaNNeo (changes requested):

imaNNeo:
And there are two things to mention:
1. We have a dash pattern feature in our line chart, so I guess the stroke should work for that case as well. (stroke around each dash section)
2. Please add some test cases and update your branch.

Thanks!

PR #2080, opened 2026-03-25
fix: Axis labels disappearing due to incorrect padding calculation in title filtering

### LINKED ISSUE #2078: Bar chart titles disappearing depending on other titles reserved space
In my app I encountered a bug with horizontal bar charts that I want to fix.
What is the bug? 
When rotating a bar chart to make it horizontal, the top titles (original "left titles") will completely disappear, as soon as the left titles (original "bottom titles") have more reserved space than the height of the chart.

**EDIT: I just realized this is not only affecting horizontal/rotated charts, but vertical charts too. It´s just less likely to have so thin charts / large reserved space there! So basically all charts should be affected?**

Luckily, it´s easy to reproduce, even in your examples. Here is the original:

Image

Since the aspect ratio is pretty evened out, I need to put in a lot of reserved space to provoke it:

Image

But if I change the aspect ratio and along with it the height of the widget, the issue appears much earlier:

Image

Since in my app, I need very slim charts with just two bars, I get this issue very, very early, with a reserved space of 60 already. This is really annoying to work around.

As far as I see, the titles disappear in side_title_widget.dart line 183:
`axisPositions = _getPositionsWithinChartRange(axisPositions, side);`
`axisPositions` has 0 items after the call of `_getPositionsWithinChartRange`.
The issue is this bit:
```
final chartSize = Size(
      widget.parentSize.width - thisSidePaddingTotal,
      widget.parentSize.height - thisSidePaddingTotal,
    ).rotateByQuarterTurns(widget.axisChartData.rotationQuarterTurns);
```
The **this side** padding is subtracted from **both** sides and with rotation, the _other_ side (that the padding is not for) is checked against. As soon as this gets negative, the titles are thrown out.

I can try to fix this and make a PR, if you want.

**Versions**
1.2.0

Enurian:
# Description

This fixes issue #2078, #1963
The box that checks if titles should be drawn was calculated incorrectly, not taking different padding on different axis (and rotation) into account. If a "reserved space" property was larger than the height/width, it made the other titles disappear.
The fix is pretty straight forward in my opinion: Take these things into account.

## Checklist

- [ ] I have updated/added relevant documentation and added dartdoc comments with `///`.

## Breaking Change?

## Related Issues

Fixes #2078 

### DIFF
--- a/lib/src/chart/base/axis_chart/side_titles/side_titles_widget.dart
+++ b/lib/src/chart/base/axis_chart/side_titles/side_titles_widget.dart
@@ -213,10 +213,19 @@ class _SideTitlesWidgetState extends State<SideTitlesWidget> {
     List<AxisSideTitleMetaData> axisPositions,
     AxisSide side,
   ) {
+    final titlesPadding = titlesData.allSidesPadding;
+    final borderPadding = widget.axisChartData.borderData.allSidesPadding;
+    final rotatedParentSize = widget.parentSize.rotateByQuarterTurns(
+      widget.axisChartData.rotationQuarterTurns,
+    );
     final chartSize = Size(
-      widget.parentSize.width - thisSidePaddingTotal,
-      widget.parentSize.height - thisSidePaddingTotal,
-    ).rotateByQuarterTurns(widget.axisChartData.rotationQuarterTurns);
+      rotatedParentSize.width -
+          titlesPadding.horizontal -
+          borderPadding.horizontal,
+      rotatedParentSize.height -
+          titlesPadding.vertical -
+          borderPadding.vertical,
+    );
     // Add 1 pixel to the chart's edges to avoid clipping the last title.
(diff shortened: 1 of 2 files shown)


imaNNeo:
Good catch. Thanks for the contribution.
As I understood, this bug has been happening since we added the `rotationQuarterTurns` property. right?

Are you able to add a test case for this fix?

Enurian:
Added a test.
No, I believe this bug was there for more than 2 years according to blame. Rotating (bar) charts just let this become an realistic issue, because you need a relatively short side. Basically every chart and every rotation is/was affected, when the ratio was not 1/1.

edit: By the way, there is a theoretical edge case that I did not test for or fix - if you set a reserved space larger than the (correct) related chart side, exceptions occur. But I guess that is user fault and would result in a stupid (not visible) chart anyways.

imaNNeo:
Nice, thanks so much!

PR #2083, opened 2026-04-16
feat: Add dot pattern for pie chart

jay-k98:
Pattern support for PieCharts was requested in  https://github.com/imaNNeo/fl_chart/issues/2082.

This Pull Request will implement this feature.

# Solution

Adding a [DotPattern] class to hold required data. Both [PieChartSectionData] and [PieChartStackSegmentData] will accept an optional pattern parameter that defaults to DotPattern.disabled().

When a section is painted, dots are drawn over the full section fill if pattern is enabled. Each stack segment is then painted on top. Segments with their own pattern get dots drawn clipped to that segment's bounds only.

-> Open for optimizations :)

# Todo's and Improvements

- Add more patterns
- Add tests
- Add documentation
- Add patterns to other chart types

### DIFF
--- a/example/lib/presentation/samples/chart_samples.dart
+++ b/example/lib/presentation/samples/chart_samples.dart
@@ -27,6 +27,7 @@ import 'pie/pie_chart_sample1.dart';
 import 'pie/pie_chart_sample2.dart';
 import 'pie/pie_chart_sample3.dart';
 import 'pie/pie_chart_sample4.dart';
+import 'pie/pie_chart_sample5.dart';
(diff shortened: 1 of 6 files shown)


imaNNeo:
Hi @jay-k98 
Thanks for your solution.
It is similar to #1958
So I would wait until we find a good solution for that.
I don't think having some for loops to have a pattern is efficient and performant.
I suggested using shaders, but I'm still not sure what's the best way to implement it.

Let's see how it goes with #1958

PR #2088, opened 2026-04-19
feat: Support per-side rod borders used for bar chart

### LINKED ISSUE #1194: Abilit to specify a border with different border sides for bar charts.
Currently BarChartRodData only takes a BorderSide and the choices are a full border or no border. 

I need to be able to specify a top border only.

haashem:
# Description

add support for per-side borders on `BarChartRodData` so users can render borders like top-only, left-only, etc.
Bar border are rendered inside - outside/centers makes bars look taller/shorter than data and can collide with nearby bars.

Resolves #1194

### API
- Added optional `Border? border` to `BarChartRodData`.
- Kept existing `borderSide` for backward compatibility.
- Rendering precedence:
  - If `border` is provided, it is used.
  - Otherwise, existing `borderSide` behavior is preserved.

### Painter updates
- Refactored bar border painting to support:
  - uniform border (`borderSide`)
  - per-side border (`border`)
  - dashed and non-dashed variants
- Added edge-based path handling for top/right/bottom/left so per-side rendering works with rounded bars as well.

Screenshot 2026-04-19 at 22 43 19

Screenshot 2026-04-19 at 22 42 27

Screenshot 2026-04-19 at 22 44 37

Screenshot 2026-04-19 at 23 03 02

## Checklist

- [ ] I have followed the [Contributor Guide] when preparing my PR.
- [ ] I have updated/added tests for ALL new/updated/fixed functionality.
- [ ] I have updated/added relevant documentation and added dartdoc comments with `///`.

## Breaking Change?

## Related Issues

Closes #1194

### DIFF
--- a/lib/src/chart/bar_chart/bar_chart_data.dart
+++ b/lib/src/chart/bar_chart/bar_chart_data.dart
@@ -339,6 +339,10 @@ class BarChartRodData with EquatableMixin {
     double? width,
     BorderRadius? borderRadius,
     this.borderDashArray,
+    @Deprecated(
+      'Use [border] instead. '
+      'This will be removed in the next major release.',
+    )
     BorderSide? borderSide,
     Border? border,
     BackgroundBarChartRodData? backDrawRodData,
@@ -389,6 +393,8 @@ class BarChartRodData with EquatableMixin {
   final List<int>? borderDashArray;
 
   /// If you want to have a border for rod, set this value.
+  ///
+  /// Deprecated: Use [border] for per-side border configuration.
   final BorderSide borderSide;
 
   /// If you want different borders for each side of rod, set this value.
@@ -423,6 +429,10 @@ class BarChartRodData with EquatableMixin {
     double? width,
     BorderRadius? borderRadius,
     List<int>? dashArray,
+    @Deprecated(
+      'Use [border] instead. '
+      'This will be removed in the next major release.',
+    )
     BorderSide? borderSide,
     Border? border,
     BackgroundBarChartRodData? backDrawRodData,
@@ -438,6 +448,7 @@ class BarChartRodData with EquatableMixin {
         width: width ?? this.width,
         borderRadius: borderRadius ?? this.borderRadius,
         borderDashArray: borderDashArray,
+        // ignore: deprecated_member_use_from_same_package, keep fallback for backward compatibility until next major.
         borderSide: borderSide ?? this.borderSide,
         border: border ?? this.border,
         backDrawRodData: backDrawRodData ?? this.backDrawRodData,
@@ -453,6 +464,7 @@ class BarChartRodData with EquatableMixin {
         width: lerpDouble(a.width, b.width, t),
         borderRadius: BorderRadius.lerp(a.borderRadius, b.borderRadius, t),
         borderDashArray: lerpIntList(a.borderDashArray, b.borderDashArray, t),
+        // ignore: deprecated_member_use_from_same_package, keep lerp fallback for backward compatibility until next major.
         borderSide: BorderSide.lerp(a.borderSide, b.borderSide, t),
         border: Border.lerp(a.border, b.border, t),
         fromY: lerpDouble(a.fromY, b.fromY, t),
--- a/lib/src/chart/bar_chart/bar_chart_data.dart
+++ b/lib/src/chart/bar_chart/bar_chart_data.dart
@@ -299,20 +299,6 @@ class BarChartGroupData with EquatableMixin {
       ];
 }
 
-Border? _normalizeBorder(Border? border, double width) {
-  if (border == null) {
-    return null;
-  }
-
-  final utils = Utils();
-  return Border(
-    top: utils.normalizeBorderSide(border.top, width),
-    right: utils.normalizeBorderSide(border.right, width),
-    bottom: utils.normalizeBorderSide(border.bottom, width),
-    left: utils.normalizeBorderSide(border.left, width),
-  );
-}
-
 /// Holds data about rendering each rod (or bar) in the [BarChart].
 class BarChartRodData with EquatableMixin {
   /// [BarChart] renders rods vertically from zero to [toY],
@@ -364,7 +350,7 @@ class BarChartRodData with EquatableMixin {
         width = width ?? 8,
         borderRadius = Utils().normalizeBorderRadius(borderRadius, width ?? 8),
         borderSide = Utils().normalizeBorderSide(borderSide, width ?? 8),
-        border = _normalizeBorder(border, width ?? 8),
+        border = Utils().normalizeBorder(border, width ?? 8),
         backDrawRodData = backDrawRodData ?? BackgroundBarChartRodData(),
         rodStackItems = rodStackItems ?? const [];
 
--- a/lib/src/chart/bar_chart/bar_chart_data.dart
+++ b/lib/src/chart/bar_chart/bar_chart_data.dart
@@ -299,12 +299,28 @@ class BarChartGroupData with EquatableMixin {
       ];
 }
 
+Border? _normalizeBorder(Border? border, double width) {
+  if (border == null) {
+    return null;
+  }
+
+  final utils = Utils();
+  return Border(
+    top: utils.normalizeBorderSide(border.top, width),
+    right: utils.normalizeBorderSide(border.right, width),
+    bottom: utils.normalizeBorderSide(border.bottom, width),
+    left: utils.normalizeBorderSide(border.left, width),
+  );
+}
+
 /// Holds data about rendering each rod (or bar) in the [BarChart].
 class BarChartRodData with EquatableMixin {
   /// [BarChart] renders rods vertically from zero to [toY],
   /// and the x is equivalent to the [BarChartGroupData.x] value.
   ///
-  /// It renders each rod using [color], [width], and [borderRadius] for rounding corners and also [borderSide] for stroke border.
+  /// It renders each rod using [color], [width], and [borderRadius] for
+  /// rounding corners and also [borderSide] or [border] for stroke
+  /// border.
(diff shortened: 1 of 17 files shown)


lib/src/chart/bar_chart/bar_chart_data.dart
```diff
@@ -299,12 +299,28 @@ class BarChartGroupData with EquatableMixin {
       ];
 }
 
+Border? _normalizeBorder(Border? border, double width) {
```
imaNNeo: I think it makes more sense to define a function in our utils to do this.
(Just like what we have for `normalizeBorderSide`, we can have another function called `normalizeBorder`.

Then we can re-use it, test it, and keep it separated.

lib/src/chart/bar_chart/bar_chart_data.dart
```diff
@@ -347,6 +364,7 @@ class BarChartRodData with EquatableMixin {
         width = width ?? 8,
         borderRadius = Utils().normalizeBorderRadius(borderRadius, width ?? 8),
         borderSide = Utils().normalizeBorderSide(borderSide, width ?? 8),
```
imaNNeo, line 356: I think we can now make borderSide deprecated. (then remove it in our next major release).
What do you think?

lib/src/chart/bar_chart/bar_chart_painter.dart
```diff
@@ -396,26 +424,344 @@ class BarChartPainter extends AxisChartPainter<BarChartData> {
...
+    required List<int>? dashArray,
+  }) {
+    if (borderSide.width <= 0 || borderSide.color.a <= 0) {
+      return;
+    }
+
+    _barStrokePaint
+      ..style = PaintingStyle.stroke
+      ..color = borderSide.color
+      ..strokeWidth = borderSide.width * 2
```
imaNNeo: This is a breaking change. In the previous version, we draw the border using `borderSide.width`, but after your change, it is `borderSide.width * 2`.

So for the users who just upgrade the version of the library, they're gonna see a doubled strokeWidth in their design

lib/src/chart/bar_chart/bar_chart_painter.dart
```diff
@@ -396,26 +424,344 @@ class BarChartPainter extends AxisChartPainter<BarChartData> {
...
+        canvasWrapper: canvasWrapper,
+        barClipPath: barClipPath,
+        barRRect: barRRect,
+        borderSide: fallbackBorderSide,
+        dashArray: barRod.borderDashArray,
+      );
+      return;
+    }
+
+    final dashArray = barRod.borderDashArray;
```
imaNNeo: It's better to define it earlier to be able to use it for both `_drawUniformBarBorderStroke()` and `_drawSolidPerSideBarBorder()`.

Now:
* `_drawUniformBarBorderStroke()` uses `barRod.borderDashArray`
* `_drawSolidPerSideBarBorder()` uses  `dashArray`

But they can just use the same `dashArray`

lib/src/chart/bar_chart/bar_chart_painter.dart
```diff
@@ -396,26 +424,344 @@ class BarChartPainter extends AxisChartPainter<BarChartData> {
...
+    _barStrokePaint
+      ..style = PaintingStyle.stroke
+      ..color = borderSide.color
+      ..strokeWidth = borderSide.width * 2
+      ..strokeCap = StrokeCap.butt
+      ..strokeJoin = StrokeJoin.miter;
+
+    final borderPath = Path()..addRRect(barRRect);
+    canvasWrapper
+      ..save()
```
imaNNeo: I'm not following why we need to have a `.save()` and `.restore()` here?
(Because before that, we didn't need them)

lib/src/chart/bar_chart/bar_chart_painter.dart
```diff
@@ -12,6 +12,34 @@ import 'package:fl_chart/src/utils/canvas_wrapper.dart';
 import 'package:fl_chart/src/utils/utils.dart';
 import 'package:flutter/material.dart';
 
+enum _BarBorderEdge { top, right, bottom, left }
+
+class _BorderEdgePathConfig {
```
imaNNeo: Can you explain a little bit why we need to have this object here?
I think I'm missing something, and here's what I got from my LLM (claude) about it:

> Painter complexity: _BorderEdgePathConfig is over-engineered
> 
>   ~200 lines of _BorderEdgePathConfig + a 4-case switch building near-identical configs is a lot of code > to draw four sides. The fundamental operation is:
> 
>   ▎ For each edge: draw the rounded-rect border path, clipped to just that edge's band.
> 
>   Which is ~15 lines with clipRect:
> ```dart
>   void _drawSolidPerSideBarBorder(CanvasWrapper cw, Path barClipPath, RRect rrect, Border border) {
>     for (final edge in _BarBorderEdge.values) {
>       final side = _borderSideForEdge(border, edge);
>       if (side.width <= 0 || side.color.a <= 0) continue;
>       cw
>         ..save()
>         ..clipRect(_borderRectForEdge(rrect, side.width, edge))  // clip to this edge's band
>         ..clipPath(barClipPath)                                   // stay inside the bar
>         ..drawPath(
>           Path()..addRRect(rrect),
>           Paint()
>             ..style = PaintingStyle.stroke
>             ..color = side.color
>             ..strokeWidth = side.width * 2,
>         )
>         ..restore();
>     }
>   }
> ```

imaNNeo (changes requested):

lib/src/chart/bar_chart/bar_chart_data.dart
haashem: Good point, Done ✅

lib/src/chart/bar_chart/bar_chart_data.dart
haashem: Totally agree. I deprecated the `borderSide`.

lib/src/chart/bar_chart/bar_chart_painter.dart
haashem: They won't see  doubled strokeWidth. With new stroke inside rendering, `* 2` is used to keep the stroke visible inside thickness around `w`, not `2w`. `clipPath(barClipPath)` will clip the outer border. so border will always be exactly `borderSide.width `

lib/src/chart/bar_chart/bar_chart_painter.dart
haashem: `save()/restore()` is required because of `clipPath()` line 493.
Previously the border was drawn directly using `drawPath()` no clipping was involved.

lib/src/chart/bar_chart/bar_chart_painter.dart
haashem: Clear and neat solution, I applied suggested solution.

Thanks a lot for your fast review.

lib/src/chart/bar_chart/bar_chart_painter.dart
imaNNeo: This is my test result on BarChartSample5 after adding this code on `bar_chart_sample5.dart:126`:
```dart
borderSide: BorderSide(
  color: Colors.black,
  width: 10
),
```

| Before (main branch) | After (your branch) |
|--------|--------|
| image | image |


Please make sure we're not introducing any breaking changes by testing your change.

haashem: The regression came from changing border rendering from centered stroke to inside-only rendering. I’ve reverted it to centered stroke behavior.

I also compared before/after tests: red is rendering on main, and blue is rendering with this change; the output matches.

In addition, I tested individual side-border rendering cases.

Thanks again for your quick reviews.

| Test 1 | Test 2 | Test 3 |
| -------- | -------- | -------- |
| border-3  | border-2  | border  |



Stroke on round rod, it looks like a hat but it is correct since stroke is rendered centered so the bleeding is expected: 
Screenshot 2026-04-21 at 22 54 19

imaNNeo:
It seems borderRadius doesn't work well with the new changes:

For example, this is our sample 1 with the below change:
```dart
BarChartRodData(
  borderRadius: BorderRadius.circular(24),
  border: Border(top: BorderSide(color: Colors.red, width: 6)),
  ...
```

image


It should work with all the edge cases, for example this is what I get from the Flutter's container widget:

```dart
Center(
      child: Container(
        width: 200,
        height: 200,
        decoration: BoxDecoration(
          color: Colors.white,
          border: Border(top: BorderSide(color: Colors.red, width: 6)),
          borderRadius: BorderRadius.all(Radius.circular(12))
        ),
      ),
    )
```
image


I think the issue is about the fact that we draw the border on the edge, but Flutter draws it inside the rect.
Let me know what you think about it

haashem:
Exactly, that was the initial reason I used internal border rendering but caused regression for current user. What I think is better to conditionally render disjointed curved borders inside. 

- Full borders: centered stroke
- Disjoint straight borders: centered stroke
- Disjoint curved borders: inside stroke only

If we want to avoid this condition, then by default all borders should be rendered inside by default, but as you highlighted in earlier comments, it cause regressions for existing users.

Or maybe we can render border on the stroke center (default) but smooth the ends:

On stroke center but smoothing the edges and with multi color borders, doesn't look valid. Flutter BoxDecoration enforces using single color for all borders.

Screenshot 2026-04-30 at 22 26 12

I have added new changes which only renders disjointed borders inside, code can be a little simplified if we use single color border. 

| All Round Borders | Partial Side Borders | Border.all | Top Border | Dashed Borders |
|----------|----------|----------|----------|----------|
| Screenshot 2026-04-30 at 22 42 43   | Screenshot 2026-04-30 at 22 42 57   | Screenshot 2026-04-30 at 22 43 30   | Screenshot 2026-04-30 at 22 43 50   | Screenshot 2026-04-30 at 22 52 55   |

haashem:
@imaNNeo May I have your review again?

PR #2092, opened 2026-04-22
feat: Add section offset to pie chart

jay-k98:
This Pull Request aims to implement the requested "exploding" pie chart sections of https://github.com/imaNNeo/fl_chart/issues/2091.

# Solution

Adding a [sectionOffset] parameter to [PieChartSectionData] and use that to calculate the offset in [drawSections] of [PieChartPainter].

### DIFF
--- a/lib/src/chart/pie_chart/pie_chart_data.dart
+++ b/lib/src/chart/pie_chart/pie_chart_data.dart
@@ -157,6 +157,7 @@ class PieChartSectionData with EquatableMixin {
     Color? color,
     this.gradient,
     double? radius,
+    double? sectionOffset,
     bool? showTitle,
     this.titleStyle,
     String? title,
@@ -169,6 +170,7 @@ class PieChartSectionData with EquatableMixin {
   })  : value = value ?? 10,
         color = color ?? Colors.cyan,
         radius = (radius ?? 40).clamp(0, double.infinity).toDouble(),
+        sectionOffset = sectionOffset ?? 0,
         showTitle = showTitle ?? true,
         title = title ?? (value == null ? '' : value.toString()),
         borderSide = borderSide ?? const BorderSide(width: 0),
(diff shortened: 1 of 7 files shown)


lib/src/chart/pie_chart/pie_chart_painter.dart
```diff
@@ -870,8 +893,14 @@ class PieChartPainter extends BaseChartPainter<PieChartData> {
       final sectionCenterAngle = startAngle + (sweepAngle / 2);
       final centerRadius = calculateCenterRadius(viewSize, holder);
 
+      final sectionOffset = section.sectionOffset;
```
imaNNeo: This logic of calculating offset is repeated multiple times in this file.
Can we just have a private shared function / getter for that instead of repeating it?

lib/src/chart/pie_chart/pie_chart_data.dart
```diff
@@ -195,6 +197,10 @@ class PieChartSectionData with EquatableMixin {
   /// Defines the radius of section.
   final double radius;
 
+  /// Additional radial translation applied to the whole section (in logical pixels).
+  /// Positive values move the section outward along its center angle.
+  final double sectionOffset;
```
imaNNeo: What do you think if we rename this new property to something like `radialOffset`?
Because commonly, we use the offset name for x and y in this project.

lib/src/chart/pie_chart/pie_chart_painter.dart
```diff
@@ -821,12 +836,20 @@ class PieChartPainter extends BaseChartPainter<PieChartData> {
...
+      final offsetDy =
+          math.sin(Utils().radians(sectionCenterAngle)) * sectionOffset;
+
       final sectionPath = generateSectionPath(
         section,
         data.sectionsSpace,
         tempAngle,
         sectionAngle,
-        center,
+        center.translate(offsetDx, offsetDy),
```
imaNNeo: And there's an issue with our rendering when we only show a single section (360 degrees).
I don't know what's the best way to handle it (as it is an edge case), but I think ignoring it is just fine!
But please explain it in the docs that if there's only one section (360 degrees), this parameter is ignored.

https://github.com/user-attachments/assets/70d5ffcc-d1a3-4701-80f2-094fc02ee0d7

imaNNeo (changes requested):

imaNNeo:
Please fix those 3 comments. After that we're good to go!

jay-k98:
Hi @imaNNeo Thanks for the feedback and yes I agree with you. I updated the branch

PR #2094, opened 2026-04-23
fix: Fix missing and incorrect copyWith parameters across chart tooltip data classes

WandsonDev:
## Description

Several `copyWith` methods across chart data classes were incomplete or had bugs that silently ignored caller-supplied values, making it impossible to update chart configuration through `copyWith`.

### Changes

**`BarChartRodData.copyWith`** — The parameter was named `dashArray` instead of `borderDashArray`, causing the constructor call to resolve to `this.borderDashArray` (the instance field) instead of the caller-supplied value. Renamed to `borderDashArray` and added the missing `?? this.borderDashArray` fallback.

**`BarTouchTooltipData`** — Added missing `copyWith` method. Also added the `direction` field to `props`, which was missing from the equality check — two objects with different `direction` values were incorrectly considered equal.

**`LineTouchTooltipData`** — Added missing `copyWith` method.

**`CandlestickTouchTooltipData.copyWith`** — Added missing `showOnTopOfTheChartBoxArea` parameter.

**`ScatterTouchTooltipData.copyWith`** — Added missing `tooltipBorderRadius` parameter.

**`ScatterChartData.props`** — Removed duplicate `rangeAnnotations` entry.

### Tests

Regression tests added for each fix to prevent regressions.

### Checklist

- [x] `make sure` passes (tests + checkstyle)
- [x] Regression tests added
- [x] No breaking changes (all additions are additive; parameter rename in `BarChartRodData.copyWith` aligns with the existing field name — the old `dashArray` name was never usable due to the bug)

### DIFF
--- a/lib/src/chart/bar_chart/bar_chart_data.dart
+++ b/lib/src/chart/bar_chart/bar_chart_data.dart
@@ -412,7 +412,7 @@ class BarChartRodData with EquatableMixin {
     Gradient? gradient,
     double? width,
     BorderRadius? borderRadius,
-    List<int>? dashArray,
+    List<int>? borderDashArray,
     BorderSide? borderSide,
     BackgroundBarChartRodData? backDrawRodData,
     List<BarChartRodStackItem>? rodStackItems,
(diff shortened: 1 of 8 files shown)


lib/src/chart/bar_chart/bar_chart_data.dart
```diff
@@ -412,7 +412,7 @@ class BarChartRodData with EquatableMixin {
     Gradient? gradient,
     double? width,
     BorderRadius? borderRadius,
-    List<int>? dashArray,
```
imaNNeo, line 415: The only thing that worries me is this property that we're removing (it adds a breaking change)
But the point is that the value here was ignored. So I don't know what to do, haha.

Maybe we can go ahead with the breaking change as it is a special edge-case

imaNNeo:
Please let me know what you think, other than that, it looks good to me.

lib/src/chart/bar_chart/bar_chart_data.dart
imaNNeo: In a professional way, we should deprecate the previous name and introduce the new name. But in this case it doesn't really make sense.

PR #2102, opened 2026-05-19
fix: Guard signingConfigs behind key.properties existence check

sunbird89629:
## Summary
- Wrap `signingConfigs` and its usage in `keystorePropertiesFile.exists()` check
- Prevents `null cannot be cast to non-null type kotlin.String` error when `key.properties` is missing

## Problem
The example app's `build.gradle.kts` unconditionally accesses `keystoreProperties["keyAlias"] as String` in the `signingConfigs` block. When `key.properties` doesn't exist (fresh clone, local dev), this crashes with:
```
null cannot be cast to non-null type kotlin.String
```

## Fix
Guard both the `signingConfigs` creation and its usage in `buildTypes.release` behind `keystorePropertiesFile.exists()`.

## Test plan
- [x] Verify `flutter build apk` works without `key.properties`
- [x] Verify `flutter build apk --release` works with valid `key.properties`

### DIFF
--- a/example/android/app/build.gradle.kts
+++ b/example/android/app/build.gradle.kts
@@ -36,18 +36,22 @@ android {
         versionName = flutter.versionName
     }
 
-    signingConfigs {
-        create("release") {
-            keyAlias = keystoreProperties["keyAlias"] as String
-            keyPassword = keystoreProperties["keyPassword"] as String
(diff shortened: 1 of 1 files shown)


imaNNeo:
Good catch, thanks for your contribution!

PR #2107, opened 2026-06-14
fix: Clip only the enabled FlClipData sides in line chart

### LINKED ISSUE #1262: Whenever any clip side is true, all sides are clipped

martin-dehlan:
# Description
`LineChart`'s `clipToBorder()` built a single rect and passed it to `clipRect`, which always clips all four sides. Disabled sides were left at the canvas edge, so they still clipped overflow. As a result, enabling any single side in `FlClipData` clipped all four sides.

Disabled sides now extend to infinity, leaving them unbounded in `clipRect`; only the sides explicitly enabled in `FlClipData` get a real boundary. (`Canvas.clipRect` accepts infinite rects — its validity check only rejects `NaN`.)

## Checklist

- [-] I have updated/added relevant documentation and added dartdoc comments with `///`.
- [-] I have updated/added relevant examples in `example`.

## Breaking Change?

## Related Issues
Closes #1262

### DIFF
--- a/lib/src/chart/line_chart/line_chart_painter.dart
+++ b/lib/src/chart/line_chart/line_chart_painter.dart
@@ -203,10 +203,13 @@ class LineChartPainter extends AxisChartPainter<LineChartData> {
     final clip = data.clipData;
     final border = data.borderData.show ? data.borderData.border : null;
 
-    var left = 0.0;
-    var top = 0.0;
-    var right = viewSize.width;
-    var bottom = viewSize.height;
+    // Sides that aren't enabled in [clip] must stay unclipped. We extend their
+    // boundary to infinity so [Canvas.clipRect] only clips the enabled sides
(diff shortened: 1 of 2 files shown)


imaNNeo:
Thanks for your contribution!

PR #2114, opened 2026-07-25
fix: Dispose gesture recognizers in RenderBaseChart

### LINKED ISSUE #1975: Memory leak detection with leak_tracker_flutter_testing
**Describe the bug**
After adding the `leak_tracker_flutter_testing` dependency to the dev dependencies to detect memory leaks while running the tests. I found two memory leaks that happens multiple times in different tests. They are not disposed objects as listed below: 
- **PanGestureRecognizer** object in `render_base_chart.dart` file
- **TapGestureRecognizer** object in `render_base_chart.dart` file
- **LongPressGestureRecognizer** object in `render_base_chart.dart` file

The above memory leaks can be fixed easily just by disposing them in the `detach` method like below: 
```dart
  @override
  void detach() {
    _validForMouseTracker = false;
    panGestureRecognizer.dispose();
    tapGestureRecognizer.dispose();
    longPressGestureRecognizer.dispose();
    super.detach();
  }
```
Furthermore, it has some other memory leaks, which are hard for me to fix as I'm not very familiar with the source code architecture.
- **TransformationController** in `axis_chart_scaffold_widget.dart` file.
- **TransformationController** in `custom_interactive_viewer.dart` file.

**To Reproduce**
- Add `leak_tracker_flutter_testing` to the dev dependancies 
```
dev_dependencies:
  ...
  leak_tracker_flutter_testing: any
```
- Create `flutter_test_config.dart` file in the root of the test folder
```
import 'dart:async';

import 'package:leak_tracker_flutter_testing/leak_tracker_flutter_testing.dart';

Future<void> testExecutable(FutureOr<void> Function() testMain) async {
  LeakTesting.enable();
  LeakTesting.settings = LeakTesting.settings
    .withIgnored(createdByTestHelpers: true);

  await testMain();
}
```
- run tests to see detected memory leaks. 

**Screenshots**
If applicable, add screenshots, or videoshots to help explain your problem.

**Versions**
 - Flutter Version: 3.35.1
 - FlChart Version: 1.0.0

I want to create a pull request to fix those memory leaks, but I can't fix the `TransformationController` leaks unless you help me.

a1573595:
# Description

`RenderBaseChart.initGestureRecognizers` creates a `PanGestureRecognizer`, a `TapGestureRecognizer` and a `LongPressGestureRecognizer`, but nothing ever released them. With the `leak_tracker_flutter_testing` harness from the issue, `line_chart_test.dart` alone reports 70 `notDisposed` leaks for those three types.

**Where they are disposed.** The issue suggests disposing in `detach()`. That would leak-check clean but is not safe: a `RenderObject` can be detached and re-attached (for example when it moves in the tree), and the recognizers must still be usable after that — disposing them in `detach()` would leave the chart with dead recognizers on re-attach. They are disposed in `dispose()` instead, which the framework calls exactly once when the render object is really gone.

**About the TransformationController leaks.** I could not reproduce them as a library bug. Enabling creation stack traces in the leak tracker showed every leaked `TransformationController` was constructed inside the test files themselves, not by `AxisChartScaffoldWidget` or `CustomInteractiveViewer` — both of those already dispose the controller they own and correctly leave an externally provided one alone. The fix here is therefore on the test side: the tests now dispose the controllers they create (via `addTearDown`). No library change was needed for that half of the report.

With the issue's harness enabled locally, the whole suite (664 tests) runs leak-free.

## Test Result

Reproduced with the harness from the issue — `leak_tracker_flutter_testing` as a dev dependency plus `test/flutter_test_config.dart`. It is already a transitive dependency of `flutter_test`, so no dependency change is needed in this repo; I added it locally only to produce the reports below and reverted it afterwards.

Both runs use the exact same command on the same test file:

```
flutter test test/chart/line_chart/line_chart_test.dart
```

### Before — `main`

23 tests pass, then `tearDownAll` fails with 70 undisposed objects:

```
notDisposed:
  total: 70
  objects:
    PanGestureRecognizer:
      test: passes canBeScaled true for FlScaleAxis.free
      identityHashCode: 764283347
    TapGestureRecognizer:
      test: passes canBeScaled true for FlScaleAxis.free
      identityHashCode: 453320791
    LongPressGestureRecognizer:
      test: passes canBeScaled true for FlScaleAxis.free
      identityHashCode: 811505688
    ... (69 recognizer entries in total)
```

Full log: [leak-report-before.txt](https://github.com/user-attachments/files/30373631/leak-report-before.txt)

### After — this branch

Same 23 tests, leak-free:

```
00:01 +23: (tearDownAll)
00:01 +23: All tests passed!
```

Full log: [leak-report-after.txt](https://github.com/user-attachments/files/30373630/leak-report-after.txt)

Running the whole suite (664 tests) with the same harness is also leak-free.

## Checklist

- [x] I have updated/added relevant documentation and added dartdoc comments with `///`.

## Breaking Change?

## Related Issues

Closes #1975

### DIFF
--- a/lib/src/chart/base/base_chart/render_base_chart.dart
+++ b/lib/src/chart/base/base_chart/render_base_chart.dart
@@ -225,4 +225,12 @@ abstract class RenderBaseChart<R extends BaseTouchResponse> extends RenderBox
     _validForMouseTracker = false;
     super.detach();
   }
+
+  @override
+  void dispose() {
+    panGestureRecognizer.dispose();
+    tapGestureRecognizer.dispose();
+    longPressGestureRecognizer.dispose();
+    super.dispose();
+  }
 }
--- a/test/chart/bar_chart/bar_chart_test.dart
+++ b/test/chart/bar_chart/bar_chart_test.dart
@@ -171,12 +171,14 @@ void main() {
 
       await tester.pumpAndSettle();
 
+      final transformationController = TransformationController();
+      addTearDown(transformationController.dispose);
       final transformationConfig = FlTransformationConfig(
         scaleAxis: FlScaleAxis.free,
         trackpadScrollCausesScale: true,
         maxScale: 10,
         minScale: 1.5,
-        transformationController: TransformationController(),
+        transformationController: transformationController,
       );
       await tester.pumpWidget(
         createTestWidget(
--- a/test/chart/base/axis_chart/axis_chart_scaffold_widget_test.dart
+++ b/test/chart/base/axis_chart/axis_chart_scaffold_widget_test.dart
@@ -526,6 +526,7 @@ void main() {
       );
 
       final transformationController = TransformationController();
+      addTearDown(transformationController.dispose);
       await pumpTestWidget(
         AxisChartScaffoldWidget(
           data: lineChartDataWithAllTitles,
@@ -1259,6 +1260,7 @@ void main() {
         final controller = TransformationController(
           Matrix4.identity()..scaleByDouble(3, 3, 3, 1),
         );
+        addTearDown(controller.dispose);
         Rect? chartVirtualRect;
         await tester.pumpWidget(
           MaterialApp(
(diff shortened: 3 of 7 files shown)


imaNNeo (approved):

imaNNeo:
Good catch!
Thanks

PR #2120, opened 2026-09-16
refactor!: Refactor chart classes to use Equatable instead of EquatableMixin

Crucialjun:
refactor!: Migrate from EquatableMixin to Equatable

# Description

Upgrades `equatable` from `^2.0.7` to `^3.0.0`, which removes `EquatableMixin` in favor of using `Equatable` directly as a mixin. All chart data classes (`BaseChartData`, `AxisChartData`, `BarChartData`, `CandlestickChartData`, `GaugeChartData`, `LineChartData`, `PieChartData`, `RadarChartData`, `ScatterChartData`, `AxisSideTitleMetaData`, and their nested classes) and the affected example samples are updated to `with Equatable` accordingly.

Also excludes generated `build/**` output from analysis in both `analysis_options.yaml` files, and refreshes `example/pubspec.yaml` (`google_fonts`, `go_router`, `package_info_plus`) and generated platform plugin files to match.

## Checklist

- [-] I have updated/added tests for ALL new/updated/fixed functionality.
- [-] I have updated/added relevant documentation and added dartdoc comments with `///`.
- [-] I have updated/added relevant examples in `example`.

## Breaking Change?

- [x] Yes, this PR is a breaking change.
- [ ] No, this PR is not a breaking change.

### Migration instructions

`equatable` is now constrained to `^3.0.0`. If your app depends on `equatable ^2.x` elsewhere, upgrade it to `^3.0.0` as well. Note that `equatable 3.0.0` also removes `runtimeType` from equality comparisons and no longer overrides `toString`, which can change equality/`toString` results for objects that mix in `Equatable` (including fl_chart's data classes).

## Related Issues


imaNNeo (approved):

imaNNeo:
Thanks for your contribution.
I updated the title to match our requirement.
There's also an analyzer issue, can you please have a look?

imaNNeo:
And I don't understand how this PR has a breaking change?
We just did an internal change, so it doesn't break any user-facing functionality.
Am I wrong?

Crucialjun:
It is a breaking change because it wont work for anyone who is using any equatable version lower than 3.0.0

imaNNeo:
> It is a breaking change because it wont work for anyone who is using any equatab

Ahh okay, makes sense!

imaNNeo:
Could you fix the `Code verification` issues?

Crucialjun:
> Could you fix the `Code verification` issues?

Let me check on it

Crucialjun:
@imaNNeo fixed

lib/src/chart/base/custom_interactive_viewer.dart
```diff
@@ -719,7 +719,7 @@ class _CustomInteractiveViewerState extends State<CustomInteractiveViewer>
         );
         _controller.duration = Duration(milliseconds: (tFinal * 1000).round());
         _animation!.addListener(_onAnimate);
-        unawaited(_controller.forward());
```
imaNNeo, line 722: Something is failing again, probably because of this removed `unawaited()` (and a couple of line further)

imaNNeo (changes requested):

imaNNeo:
And it seems in [equatable 3.0.0](https://pub.dev/packages/equatable/changelog#300), it doesn't compare `runtimeType` anymore.
So now these two objects are not equal (which breaks our animation functionality):
```
final circle = FlDotCirclePainter(color: Colors.red, radius: 4, strokeWidth: 1);
final square = FlDotSquarePainter(color: Colors.red, size: 4, strokeWidth: 1);
```

So I would like to ask you to add the below test in our `axis_chart_data_test.dart`:
```dart
group('FlDotPainter equality check', () {
  test('different dot painter subtypes with identical field values are not equal', () {
    final circle = FlDotCirclePainter(color: Colors.red, radius: 4, strokeWidth: 1);
    final square = FlDotSquarePainter(color: Colors.red, size: 4, strokeWidth: 1);
    final cross = FlDotCrossPainter(color: Colors.red, size: 4, width: 1);

    expect(circle, isNot(equals(square)));
    expect(circle, isNot(equals(cross)));
    expect(square, isNot(equals(cross)));
  });
});
```

And it should fail, to fix that, we need to add the `runtimeType` property to those classes.
For example in FlDotCirclePainter:
```dart
@override
List<Object?> get props => [
  runtimeType,
  color,
  radius,
  strokeColor,
  strokeWidth,
];
```

lib/src/chart/base/custom_interactive_viewer.dart
Crucialjun: What is failing ??

imaNNeo: Failed again, you can check it [here](https://github.com/imaNNeo/fl_chart/actions/runs/35504928886/job/106124452550?pr=2120)

imaNNeo:
Nice fix in axis_chart_data.dart. But the same gap still exists elsewhere — these sibling-subclass pairs also lost runtimeType in their props and only avoid colliding today by accident (different prop-list lengths):

- TouchedSpot → BarTouchedSpot, RadarTouchedSpot
- GaugeTickPainter → GaugeTickLinePainter, GaugeTickCirclePainter
- GaugePointerPainter → GaugePointerNeedlePainter, GaugePointerCirclePainter

Can we add runtimeType to these 6 props getters too, for consistency?

